Issue #10, opened 2020-02-02
Choose a method for local settings


rschroll:
We should have a way to override the default settings with local options.  This will allow:
- Developers to use different database options
- A testing environment to use different options
- Production secrets to not be committed to the repo

It looks like Django has a dozen different ways to do this: https://code.djangoproject.com/wiki/SplitSettings.  It looks like the two general ideas are
1. Define defaults in settings.py, and then import a local file, if it exists, to override defaults.
2. Have a local file import the defaults from settings.py, and then tell Django to use this other file for the settings with something like `$DJANGO_SETTINGS_MODULE`.

The former is slightly more foolproof (no forgetting the env var), but the latter gives a little more flexibility -- users can keep their local settings wherever they like.  I don't really have a strong preference.

rschroll:
We've gone with a local_settings.py file.

Issue #9, opened 2020-02-02
Specify Python version


rschroll:
To avoid issues in development, we ought to choose a minor version of Python to focus on.  At the very least, a minimum version will let us know which features we can use.  But a specific version will help ensure that the versions in `requirements.txt` will work.  (I couldn't get `mysql-connector-python` 1.1.6 installed on my system, which may be due to using Python 3.7.)

aaronoppenheimer:
I'm fine with using 3.7 - I'm on 3.8 for no real reason. 3.7.6 is latest stable 3.7 - good with that?

rschroll:
3.7 is fine with me, but I'm also happy to move to 3.8.  No reason not to join the future.

aaronoppenheimer:
Well, then, future it is. Now we get the walrus operator. Just what I've been waiting for.

PR #11, opened 2020-02-02
Set up CI testing via Github actions

rschroll:
This adds an action to run the Django tests on every push.  To make this easier to set up, I switched the default database to use SQLite.  Depending on what we do in #10, this may or may not be the right way to handle things.

Right now, tests are run on Python 3.7, only because that's what I've been using.  Once we pick a preferred version (see #9), we can adjust that here.  Alternatively, Github actions make it relatively easy to run the same tests against multiple Python versions.


rschroll:
The action ran on my account, but it didn't link the result to this PR, for whatever reason.  You can see the output here: https://github.com/rschroll/GO3/runs/422149229.

var/.gitignore
```diff
@@ -0,0 +1 @@
+*
```
aaronoppenheimer: Doesn't this just silence git from reporting anything that isn't added?

aaronoppenheimer: huh - actually, it doesn't look like anything is different.

aaronoppenheimer:
That is cool - it's almost like real devops! Thanks for setting it up.

I'm fine with using sqlite by default, and I think using a local settings override file ('from settings_local import *') is best just because it's simplest.

aaronoppenheimer:
I've pushed a change to settings.py to import settings_local.py, if it exists. works like a champ.

I'll now stop developing directly into the master branch. Right now it's set up so that I can merge without requiring a review - if I do that, do you get notified?

A

rschroll:
I haven't gotten notifications so far, but that's probably because I just joined the project.  I'm sure there's a notification setting where I can tweak it to my heart's content.  You keep up whatever workflow works for you; I'll let you know if it creates any problems on my end.

Re the .gitignore in var: A trick I've used elsewhere is to use a var folder for all sorts of files that vary from developer to developer.  That's where I've stuck the default DB, for example.  By putting the .gitignore file in the directory, rather than adding an entry to the main copy, it ensure that this directory gets created on cloning the repo.  If we were to allow multiple settings files, sticking them in var would make sense.  But the current solution doesn't admit that, so I think the current location makes sense.

aaronoppenheimer:
Ah, I see - I didn't notice var was added. Makes sense.

PR #16, opened 2020-02-10
Task queue and email sending with Django-Q

rschroll:
An experiment in setting up Django-Q and using it to send email asynchronously.  The Django-Q side was pretty straightforward.  On the email side, I guessed at an architecture that might be useful.  The basic idea is to use Django templates to create Markdown documents, which then produce both the HTML and plain text content of emails.  If the first like of this document begins with 'Subject:', it is taken to be the subject line of the email.  I can already see places where this will probably be insufficient, but I thought I should check in to see if this is going on the right track first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,3 +21,11 @@ At this point, you should be able to run the project locally:
 python manage.py runserver
 ```
 You can log in with the user created above.
+
+## Task queue
+
+Django-Q is used as a task queue, and the default setting uses the standard database as a broker.  Tasks can be serialized into the database without any additional configuration.  However, a separate process is used to run these items in the queue.  This can be launched with
+```
+python manage.py qcluster
+```
+You can cause tests to run synchronously by setting `'sync': True` in the `Q_CLUSTER` settings.
--- a/go3/settings.py
+++ b/go3/settings.py
@@ -62,6 +62,7 @@ INSTALLED_APPS = [
     'django.contrib.staticfiles',
     'django_extensions',
     'django.contrib.humanize',
+    'django_q',
 ]
 
 MIDDLEWARE = [
@@ -168,6 +169,17 @@ STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
 # Redirect to home URL after login (Default redirects to /accounts/profile/)
 LOGIN_REDIRECT_URL = '/'
 LOGOUT_REDIRECT_URL = '/accounts/login'
+
+# Configure Django-q message broker
+Q_CLUSTER = {
+    'name': 'DjangORM',
+    'workers': 4,
+    'timeout': 90,
+    'retry': 120,
+    'queue_limit': 50,
+    'bulk': 10,
+    'orm': 'default',
+}
 
 try:
     from .settings_local import *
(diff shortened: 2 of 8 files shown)


rschroll:
Two things I wasn't sure about:
- I didn't see a clear place for the email code to go, so I created `lib/` to hold these sorts of functions. I don't know what the usual Django solution is.
- I added additional packages directly to requirements.txt, rather than running `pip freeze` to capture the full state of my environment.  I don't know if we're intending this to capture just the critical packages or all of them.

aaronoppenheimer:
a) cool!

b) I think the django answer for where to put the code would be to set up a separate django "app" to hold all of the email code, with 'startapp' - that gives us the usual structure for admin, settings, migrations etc.

c) using markdown is a great idea. I don't know if anyone is actually using the raw text version of the email - will those people hate the embedded link syntax? Maybe it doesn't matter - they already have to cut-and-paste the url.

Today, the email templates are stored in the strings database and just use the standard internationalization system - that's next on my list to get working.

A

rschroll:
On Mon, Feb 10, 2020 at 4:24 AM aaronoppenheimer <notifications@github.com>
wrote:

> b) I think the django answer for where to put the code would be to set up
> a separate django "app" to hold all of the email code, with 'startapp' -
> that gives us the usual structure for admin, settings, migrations etc.
>

This was my first idea, but as I started to implement it, it felt silly.
There's no models, no views, no administration needed here. It's just some
useful functions.  If these were going to be used only from one particular
app, then we should just stick it in that app.  Perhaps it should go in
members, since members get emailed?  But I imagine that we'll want to email
whole bands, so that doesn't seem right.  To me, a lib folder that other
apps can use seemed the most straightforward.  But I'm happy to be
convinced otherwise.

>
> c) using markdown is a great idea. I don't know if anyone is actually
> using the raw text version of the email - will those people hate the
> embedded link syntax? Maybe it doesn't matter - they already have to
> cut-and-paste the url.
>

My guess is that anyone who insists on seeing the raw text email today is
enough of a geek that they can read Markdown.  (Related -- I ran into a
problem recently where some markdown was not being rendered, and I kept
finding new places where the raw markdown was peeking through.  I think I'm
so used to reading it that my brain was just rendering it for me....)

>
> Today, the email templates are stored in the strings database and just use
> the standard internationalization system - that's next on my list to get
> working.
>

It looks like Django templates support i18n:
https://docs.djangoproject.com/en/3.0/topics/i18n/translation/#internationalization-in-template-code.
I'd fear that this could get tedious for large webpages.  But since
Markdown is so readable, we can probably just mark the whole email in a
single translation block.

Robert

rschroll:
Upon further reflection, I realized that I was conflating concerns.  While sending emails shouldn't know anything about Members, preparing emails is likely to be intimately connected to the Member object.  So I split the code into two parts: A very short bit of code in lib takes email objects and sends them asynchronously.  Then, member.helpers gains a prepare_email function that worries about the templating.  Because it gets a member as input, it can translate the email into the member's preferred locale.

The way I'm imagining this working is that a band or gig would generate a list of Members who need to get an email.  We would map prepare_email over that list to get a list of emails to send, and pass that to the new send_messages_async function.  Let me know if this seems sensible.

aaronoppenheimer:
Thanks for this - it'll take a me a day or two to get my head around it and I'll get back to you on thoughts.

The i18n system is basically the same as I was using before - hopefully the language files will just work!

PR #25, opened 2020-02-16
Send new gig email on gig creation

rschroll:
It's still in a rough form, but the basics work.  Hopefully this is enough to see if we're on the right track or not.  A few comments:

- There's a fake get_confirm_urls function as a placeholder, since I don't see anything around this functionality currently implemented.
- Much of this should generalize to also handle reminder emails and change emails, but I haven't done any work in that direction.  I'd rather see how this develops before figuring out the right places to generalize.
- A lot of the display logic has been moved into the template itself.  Since the template rendering is happening in the member's locale, this means we can get dates and times automatically localized, without functions like format_date_for_member.  The downside is that the Django templates don't seem to have an option to control whitespace before or after tags (as Jinga templates do), making the template a bit difficult to read.  We might be able to address this with some custom filters, but it's not all that bad as-is.

### DIFF
--- a/gig/helpers.py
+++ b/gig/helpers.py
@@ -59,3 +59,6 @@ def update_plan_default_section(assoc):
     the default section of the member assoc has changed, so update any plans that aren't overriding
     """
     Plan.objects.filter(assoc=assoc, plan_section=None).update(section=assoc.default_section)
+
+def get_confirm_urls(member, gig):
+    return {'yes_url': 'http://example.com/yes', 'no_url': 'http://example.com/no', 'snooze_url': 'http://example.com/snooze'}
--- a/gig/helpers.py
+++ b/gig/helpers.py
@@ -18,7 +18,9 @@
 from django.http import HttpResponse
 from django.contrib.auth.decorators import login_required
 from .models import Plan
-from band.models import Section
+from band.models import Section, AssocStatusChoices
+from member.helpers import prepare_email
+from lib.email import send_messages_async
 
 @login_required
 def update_plan(request, pk, val):
@@ -62,3 +64,20 @@ def update_plan_default_section(assoc):
 
 def get_confirm_urls(member, gig):
     return {'yes_url': 'http://example.com/yes', 'no_url': 'http://example.com/no', 'snooze_url': 'http://example.com/snooze'}
+
+def email_from_plan(plan, template):
+    gig = plan.gig
+    member = plan.assoc.member
+    contact_name, contact_email = ((contact.display_name, contact.email)
+                                   if (contact := gig.contact) else ('??', None))
+    context = {
+        'gig': gig,
+        'contact_name': contact_name,
+        **get_confirm_urls(member, gig)
+    }
+    return prepare_email(member, template, context, reply_to=[contact_email])
+
+def send_emails_from_plans(plans_query, template):
+    contactable = plans_query.filter(assoc__status=AssocStatusChoices.CONFIRMED,
+                                     assoc__email_me=True)
+    send_messages_async(email_from_plan(p, template) for p in contactable)
--- a/gig/helpers.py
+++ b/gig/helpers.py
@@ -74,7 +74,11 @@ def email_from_plan(plan, template):
                                    if (contact := gig.contact) else ('??', None))
     context = {
         'gig': gig,
+        'change_string': 'FIXME: Figure out what changed',
         'contact_name': contact_name,
+        'status': plan.status,
+        'status_label': Plan.StatusChoices(plan.status).label,
+        **Plan.StatusChoices.__members__,
         **get_confirm_urls(member, gig)
     }
     return prepare_email(member, template, context, reply_to=[contact_email])
--- a/gig/helpers.py
+++ b/gig/helpers.py
@@ -81,3 +81,7 @@ def send_emails_from_plans(plans_query, template):
     contactable = plans_query.filter(assoc__status=AssocStatusChoices.CONFIRMED,
                                      assoc__email_me=True)
     send_messages_async(email_from_plan(p, template) for p in contactable)
+
+def send_reminder_email(gig):
+    undecided = gig.member_plans.filter(status__in=(Plan.StatusChoices.NO_PLAN, Plan.StatusChoices.DONT_KNOW))
+    send_emails_from_plans(undecided, 'email/gig_reminder.md')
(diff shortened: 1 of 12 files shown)


rschroll:
I've refactored things again, but I feel like I'm converging on a solution.  Now the gig email system is based on plans (or actually a query against the plans table), rather than gigs.  This is necessary for the reminder system and the snooze system, so that we send reminders only to people who need them.  It also makes it possible to include current plan status in edit emails.  The only one it doesn't immediate fit into is the new gig email, but we're generating all the members' plans at that point anyway, so why not use them?

I've hooked up some functions to send these various types of emails.  Since the content is very similar, I've used template inheritance to avoid duplication.

The big unknown at this point is how to get the information about what's changed into the gig edit email.  I see three components to this problem:
1. How do we know what's changed in the model?  There's a whole bunch of ideas [here](https://stackoverflow.com/questions/1355150/when-saving-how-can-you-check-if-a-field-has-changed), but I think the general possibilities are
    a. Handle the POST that edits the gig manually, and check whether the incoming fields match the existing attributes.
    b. Use a [FieldTracker](https://django-model-utils.readthedocs.io/en/latest/utilities.html#field-tracker) on the Gig model so we can see what's changed after we adjust the fields (but before we save it).
    c. Use [Form.changed_data](https://docs.djangoproject.com/en/dev/ref/forms/api/#django.forms.Form.changed_data) to get an accounting of what changed in form submission.
    d. Do some sort of hackery to the Gig class to do this bookkeeping ourselves.

2. How do we get that information to the start of the email process, which is currently in a post_save signal handler?  We can't pass additional arguments into this handler.  We could
    a. Save the change information to a special field in the Gig model.
    b. Call the email process directly from an overridden Gig.save, instead of relying on signals.  (I've seen some advice that Django signals should be avoided, and the solution to (1) likely involves some alteration of the save method, so this seems reasonable.)

3. How to get the change information all the way to the template to be formatted into the email itself?  The plan-based system means that we're looking up the gig from each plan, not using a single copy of the gig that we could alter in some way.
    a. If we save the change information to the model (2a), it will already be along for the ride.  I might worry that this change information will stick around even after these emails are sent out, which could potentially cause confusion.
    b. We could pass this information through the several layers.  This feels like it's mixing up the abstraction a bit, though, since these things are supposed to take an arbitrary collection of plans.  It'd be odd for them to get an argument connected to one particular gig.
    c. We could have a general context argument that gets passed through into the template rendering.  Fundamentally, this has the same problem as (b), but since it's for general variables, somehow it doesn't feel as wrong to implement.

All of this is mostly to try to organize my thoughts, but I welcome any insight you have.

aaronoppenheimer:
Hi Robert,

I've seen the advice against using signals too - I'll go back in and look at how I used them. Overriding save methods might just be cleaner.

Keeping the last change with the gig (2a) is attractive because then we can display that information on the good detail page in a "here's what changed last, when, by whom" field. Once in a while users ask for a full-blown change log, which I think is Overkill, but keeping the last change around would be useful.

The field tracker is cool - I hadn't run into it.

A

rschroll:
A few random thoughts:

Another option for step 1 (let's call it e): Attach to the pre-save signal, where we'll get a copy of the instance to be committed.  There, we can pull the corresponding row in the database to see what's changed.  Unfortunately, we can't send the emails right away -- we need to wait until after it's saved, so we get the new gig info from the plans.  But stuffing that process entirely into async is probably enough.

As for signals, I'm not sure I buy the arguments against them.  The two I've seen are:
1. They make it code more indirect, to it's harder to see what happens in what order.  That's true, but the benefit is that we can separate code by concerns.  Sticking everything into the Gig.save method means that the Gig has to know about sending emails.  We could fight that by passing a generic function into Gig to be called at the appropriate time, but then we're just reinventing signals (poorly).  In this trade off, I generally lean towards more separation, even if it leads to more indirection.
2. The signals are handled synchronously, which makes is possible to unintentionally write slow code.  But I'm suspecting we could write a async_receiver decorator that would get a signal and immediately queue a task to be handled later.

The idea of keeping a full history of changes intrigued me.  I poked around, and I found that it's a whole thing: https://djangopackages.org/grids/g/model-audit/.  Judging by popularity, there's two front-runners.  Django-revision seems to be more sophisticated, but I found the documentation hard to follow.  Django-simple-history seemed simpler (appropriately enough), and the documentation was more readable.  Both offer the ability to keep history of only certain objects, so we're not committing to versioning the full database with this.  Even if keeping the full history is overkill, it might be easier to use an existing solution than trying to roll our own minimal solution.

Whether we use these or something else, it sounds like we're okay keeping some amount of history in the DB.  That really solves steps (2) and (3), and step (1) can wait until we figure out what the edit page looks like.  (Or, if we use an automatic history system, we might not need anything explicitly there.)

rschroll:
This is getting a little off-track, but since we were talking about signals, here's a little experiment.  It creates a `async_receiver` decorator that causes the handler to run as an async task.  I haven't tested it beyond this one case, but it could be a good way to use signals without clogging up the main process.

https://github.com/SecondLiners/GO3/commit/7a5b01672ba180c6764fbe44eb2fab9fac3880fe

aaronoppenheimer:
Signals - I can see the value in the way I've used them with the member model - where are a couple of housekeeping things that need to happen in the background that are sort of structural - so making sure there's a member preferences object for every member. But for how I used them with the gig model, where I'm more likely to scratch my head at some point trying to figure out where the default plans for each member are created without remembering it's in a post_save signal is more confusing. But I guess as long as we just keep that stuff simple, it's the same whether it's in a save function or the signals.

History - that simple-history module looks pretty great. For now, I'm only thinking about tracking gigs, which change seldom if at all, so we won't be filling the database with junk. The "history diffing" functionality seems like exactly what we want. I can play with it unless you get there first!

aaronoppenheimer:
Looks great to me. I'm running on a mac and there's a known problem with django-q but it works fine if I run it synchronously.

Breaking the templates into component strings to translate instead of a monolithic block is fine - I'll extract the rest of the useful text from the old translation files.

aaronoppenheimer:
Oh, regarding the 'yes, I can be there' etc. links in the emails - the way those are created in go2 was to encrypt the member id, gig id, and answer (yes, no, maybe) into a URL. (In retrospect I could have just used the plan id, but whatever.) Is that still a decent way to do this, or is there something better?

A

PR #29, opened 2020-02-22
Add async_receiver decorator, to trigger async task on signals

rschroll:
Here was a little experiment I made in #25.  It creates an `async_receiver` decorator that causes an async task to run on signal reception.  For testing purposes, I stuck it on the `notify_new_gig` handler.  This leads to two questions:

1. Should `notify_new_gig` run async?  The really slow part, email sending, is already in an async task.  But I could imagine that prepping all the emails for a large band could bog down too.  There's no need to do it synchronously, so I think it's worthwhile.

2. Do we need a new decorator for this?  If this is the only place we need it, it'd be simpler to just trigger an async task by hand.  But if this becomes a common pattern, the decorator would be useful.  I don't really know what to expect in the future development.

### DIFF
--- a/gig/signals.py
+++ b/gig/signals.py
@@ -18,6 +18,7 @@ from django.db.models.signals import post_save, pre_save
 from django.dispatch import receiver
 from .models import Gig, Plan
 from gig.helpers import send_emails_from_plans
+from lib.asyncsignal import async_receiver
 from datetime import datetime
 from django.utils import timezone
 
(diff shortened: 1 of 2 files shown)


aaronoppenheimer:
I doubt we'll see too many cases like this one, so I wouldn't bother getting the decorator ready for prime time.

I would definitely move the email preparation into it's own asynchronous task, though - your're right, some of the bands have hundreds of members and interpreting the templates will take time.

A

rschroll:
See #31.

PR #30, opened 2020-02-23
Better datetime

aaronoppenheimer:
Moved gig times to use three datetime objects (date [call time], set time, end time) - this lets them all be timezone aware. Added signal to ensure the set time and end times are populated. Updated email template to make sure it's working with new format.


gig/forms.py
```diff
@@ -26,9 +28,29 @@ def __init__(self, **kwargs):
...
+        date = self.cleaned_data['date']
+        if date < timezone.now():
+            raise ValidationError(_('Date must be in the future'), code='invalid date')
+        return date
+
+    def clean_setdate(self):
+        date = self.cleaned_data['date']
+        setdate = self.cleaned_data['setdate']
+        if setdate and setdate < date:
+            raise ValidationError(_('Set date must be later than the start date'), code='invalid end date')
```
rschroll: This may be confusing to users, since the set date can be the same as the start date; it's just the time that has to be later.  (Or actually, not earlier.)  Perhaps, "Set date must not be earlier than start date"?

In the emails the date is referred to as the "call", not the "start".  I don't know if that's a consistent name or not.

gig/forms.py
```diff
@@ -26,9 +28,29 @@ def __init__(self, **kwargs):
...
+        return date
+
+    def clean_enddate(self):
+        date = self.cleaned_data['date']
+        enddate = self.cleaned_data['enddate']
+        if enddate and enddate < date:
```
rschroll: Is it worth assuring that the enddate is after the setdate as well?

templates/email/gig.md
```diff
@@ -5,7 +5,7 @@ Subject: {% block subject %}{% endblock %}
 
 {{ gig.title }}
 {% trans "Date" %}: {{ gig.date|date:"SHORT_DATE_FORMAT" }} ({{ gig.date|date:"D" }}){% if gig.enddate %} - {{ gig.enddate|date:"SHORT_DATE_FORMAT" }} ({{ gig.enddate|date:"D" }}){% endif %}
-{% trans "Time" %}: {% if gig.calltime %}{{ gig.calltime|time:"TIME_FORMAT" }} ({% trans "Call Time" %}){% if gig.settime or gig.endtime %}, {% endif %}{% endif %}{% if gig.settime %}{{ gig.settime|time:"TIME_FORMAT" }} ({% trans "Set Time" %}){% if gig.endtime %}, {% endif %}{% endif %}{% if gig.endtime %}{{ gig.endtime|time:"TIME_FORMAT" }} ({% trans "End Time" %}){% endif %}
+{% trans "Time" %}: {% if gig.date %}{{ gig.date|time:"TIME_FORMAT" }} ({% trans "Call Time" %}){% if gig.setdate or gig.enddate %}, {% endif %}{% endif %}{% if gig.setdate %}{{ gig.setdate|time:"TIME_FORMAT" }} ({% trans "Set Time" %}){% if gig.enddate %}, {% endif %}{% endif %}{% if gig.enddate %}{{ gig.enddate|time:"TIME_FORMAT" }} ({% trans "End Time" %}){% endif %}
```
rschroll: Since `setdate` and `enddate` will always exist, these tests are no longer meaningful.  Perhaps we should change the test to notice if they're different from `date`?

aaronoppenheimer:
I guess you're right about setting the set and end dates automatically. I did this because I need to query on end date for the agenda page - I want to show all gigs for which the end date hasn't passed yet. I suppose I can just do that with a smarter query, and you're right that it's better to know whether the set and end dates have been explicitly set.


On Sat, Feb 22, 2020, at 7:57 PM, Robert Schroll wrote:
> *@rschroll* commented on this pull request.

> In terms of the implementation, it looks good. I'm a little worried about the provisioning of default `setdate` and `enddate` values. I'd think that users would prefer to know that these haven't been set, rather than getting a value and trying to work out if it was chosen purposefully or automatically.

> 
> In gig/forms.py <https://github.com/SecondLiners/GO3/pull/30#discussion_r382953462>:

> > @@ -26,9 +28,29 @@ def __init__(self, **kwargs):
             **kwargs
         )
 
+    def clean_date(self):
+        date = self.cleaned_data['date']
+        if date < timezone.now():
+            raise ValidationError(_('Date must be in the future'), code='invalid date')
+        return date
+
+    def clean_setdate(self):
+        date = self.cleaned_data['date']
+        setdate = self.cleaned_data['setdate']
+        if setdate and setdate < date:
+            raise ValidationError(_('Set date must be later than the start date'), code='invalid end date')
> 
> This may be confusing to users, since the set date can be the same as the start date; it's just the time that has to be later. (Or actually, not earlier.) Perhaps, "Set date must not be earlier than start date"?

> In the emails the date is referred to as the "call", not the "start". I don't know if that's a consistent name or not.

> 
> In gig/forms.py <https://github.com/SecondLiners/GO3/pull/30#discussion_r382953487>:

> > +        date = self.cleaned_data['date']
+        if date < timezone.now():
+            raise ValidationError(_('Date must be in the future'), code='invalid date')
+        return date
+
+    def clean_setdate(self):
+        date = self.cleaned_data['date']
+        setdate = self.cleaned_data['setdate']
+        if setdate and setdate < date:
+            raise ValidationError(_('Set date must be later than the start date'), code='invalid end date')
+        return date
+
+    def clean_enddate(self):
+        date = self.cleaned_data['date']
+        enddate = self.cleaned_data['enddate']
+        if enddate and enddate < date:
> 
> Is it worth assuring that the enddate is after the setdate as well?

> 
> In templates/email/gig.md <https://github.com/SecondLiners/GO3/pull/30#discussion_r382953703>:

> > @@ -5,7 +5,7 @@ Subject: {% block subject %}{% endblock %}
 
 {{ gig.title }}
 {% trans "Date" %}: {{ gig.date|date:"SHORT_DATE_FORMAT" }} ({{ gig.date|date:"D" }}){% if gig.enddate %} - {{ gig.enddate|date:"SHORT_DATE_FORMAT" }} ({{ gig.enddate|date:"D" }}){% endif %}
-{% trans "Time" %}: {% if gig.calltime %}{{ gig.calltime|time:"TIME_FORMAT" }} ({% trans "Call Time" %}){% if gig.settime or gig.endtime %}, {% endif %}{% endif %}{% if gig.settime %}{{ gig.settime|time:"TIME_FORMAT" }} ({% trans "Set Time" %}){% if gig.endtime %}, {% endif %}{% endif %}{% if gig.endtime %}{{ gig.endtime|time:"TIME_FORMAT" }} ({% trans "End Time" %}){% endif %}
+{% trans "Time" %}: {% if gig.date %}{{ gig.date|time:"TIME_FORMAT" }} ({% trans "Call Time" %}){% if gig.setdate or gig.enddate %}, {% endif %}{% endif %}{% if gig.setdate %}{{ gig.setdate|time:"TIME_FORMAT" }} ({% trans "Set Time" %}){% if gig.enddate %}, {% endif %}{% endif %}{% if gig.enddate %}{{ gig.enddate|time:"TIME_FORMAT" }} ({% trans "End Time" %}){% endif %}

rschroll:
Does Django have some sort of computed field?  If so, that could be used for the "has it passed" logic, while we use the entered (or not) date for display logic.

aaronoppenheimer:
Well, since I don't need to order by the end date, I don't really need it stored in the database. I can query "future plans" with a "Q object" that lets you do more complicated queries. I will still sort them by the start date anyway.

PR #32, opened 2020-02-23
Provide the answer links for emails

### LINKED ISSUE #28: Plan for the status links
As mentioned at the end of #25, we need figure out how the links in the gig emails to set status will work.  I really like how the existing system works.  Being able to set a status without needing to log in makes the system usable for people who are unable or uncomfortable using the web interface.  I think the first thing we need to decide is, should existing links keep working after we switch over to the new system.  I see three possibilities:

**Yes, forever.**  In this case, we'll need to match up primary keys in the old and new databases, and we might as well just keep the existing URL structure.

**Yes, but only for future gigs in the old database.** This would allow a seamless transition, but requires potentially two separate systems to handle links.  I'm not opposed to that, though, since it makes it easier to figure out which part we can remove once all these gigs have moved into the past.

**No.** In practice, I imagine that most people will click that link in the email in the first day after receiving the email or not at all.  It may be worth annoying people with broken links for that day or two of the transition to not have to worry about ensuring continuity.

If we go with either of the latter two, I like the idea of hooking it to the plan, instead of the user and gig.  We already have the plan when sending the email, and that's what we'd need to record their response.  This might also obviate the need to encrypt the keys.  I assume that was done so that someone couldn't come up with gig and member keys from separate sources and change someone else's response.  If we identify everything by the plan key, that attack is mitigated.  It would require the plan key to not be guessable, though, and in my testing, it looks like Django + SQLite just increments the primary key for each new row.  If we can get the database to give us pseudo-random primary keys, we can just use that.  Otherwise, I think we do need encryption.

rschroll:
The plans are switched over to using UUIDs for primary keys, so that they can be exposed in URLs.  Note that, while the migration will upgrade the database, it will not upgrade entries in the database.  Thus, you will encounter run-time errors if there are any plans with the old IDs in your DB.  Presumably there would be a way to do this migration if we cared, but there's no point right now.

The snooze_until logic came from the previous version, although I rewrote it a bit.  I think it does the same thing, but please check me.

I added a feature in the plan pre-save hook to unset snooze_until if a non-unsure status is chosen.  This raised another question: Should setting the status to "Don't know" automatically trigger the snooze_until being set, or should it only happen when you visit an email link?  Right now, it does the latter, since that's how the old code work.  But I can see an argument to doing it all of the time.  This would simplify the code a little bit, since it could just reside in the pre-save hook.

I also added a link to the gig info page on the answer page, because that's something I've wanted for a little while.

Since the snooze code relies on gig dates, this will be affected by the changes in #30.  I'll plan to rebase this onto that once it merges.

Closes #28.


aaronoppenheimer:
re: automatically sending reminder emails - the gig-o has a long-standing policy of "no automatic emails" but maybe there's some version that would work. Maybe when you first register your attendance, instead of "don't know" the option is "remind me later" like in the email, but it still registers as "don't know" so when you come back it works as before..?

rschroll:
Looks good to me.  I don't know if you want to merge this in now, or wait until the date management branch gets in.

aaronoppenheimer:
I merged it all in - seems all good. Thanks!

PR #34, opened 2020-02-25
Abstract gig

aaronoppenheimer:
Changed the "AbstractGig" class to be an actual abstract class, so that we can have other gig-like objects.

Will require a complete database nuking - I trashed all the old migrations.


gig/models.py
```diff
@@ -74,8 +74,13 @@ def __str__(self):
...
 
 class AbstractGig(models.Model):
+
+    class Meta:
+        abstract = True
+
     title = models.CharField(max_length=200)
-    band = models.ForeignKey(Band, related_name="gigs", on_delete=models.CASCADE)
+    band = models.ForeignKey(Band, on_delete=models.CASCADE)
+    # must access by band.gig_set - if we use a custom related_name things get difficult
```
rschroll: I think we could keep using `gigs` as the attribute name.  We'd just have to set `related_name='%(class)ss'`, I think.  See here: https://docs.djangoproject.com/en/3.0/topics/db/models/#abstract-related-name

aaronoppenheimer: I tried that and it gave me a problem I didn't understand. It's preferable, though - I'll try again.

Re: migrations - it doesn't matter to me one way or the other. I've been impressed with how few times I've had to completely trash the database. If we stop tracking them, the plan would be to just rebuild whenever necessary, and at some point declare "now!" and start checking in the migration files?

aaronoppenheimer:
OK! I don't know what I did before that didn't work but it's working now.

PR #35, opened 2020-02-29
Track history of Gigs, and explain changes in edit emails

rschroll:
This uses django-simple-history to track the changes in the Gig objects.  When a change email is sent, it uses the diff between the latest and previous versions to populate some change information.

The logic to figure out what to display is a bit involved, so most of it sits in Python, not the template.  This in turn means the translation has to be delayed until we get into the translation context, so there are a few tricks to ensure lazy evaluation.

Change emails will be triggered by any change to a gig object, even something that is not included in the email (dress, for instance).  As it's set up now, a change to status or any of the times is reported, along with the current and past values.  Everything else is reported as a change to the "details", even if it's not the details property.  What's the right way to go about this?  We could show the diff for each possible field, but we'd need to code up each possibility.  We could only send the email for fields in the email.  Or we could have an "other details have changed, see the website for them" message on these other fields.


gig/helpers.py
```diff
@@ -64,14 +68,48 @@ def update_plan_default_section(assoc):
     """
     Plan.objects.filter(assoc=assoc, plan_section=None).update(section=assoc.default_section)
 
+def date_format_func(dt, fmt):
+    # Returns lambdas that the template will evaluate in the correct
+    # language context
+    return lambda: date_format(template_localtime(dt), fmt)
+
+def date_diff(latest, previous):
+    if latest.date() == previous.date():
```
aaronoppenheimer: There might not be a previous date - enddate is 'none' by default, and if you add one later this fails

rschroll: Good catch!  This is fixed, and tested, now.

rschroll:
I've added contact as being separately called out, since that's the other piece of information actually in the email.  I'm happy lumping everything else in as details, at least right now.

PR #38, opened 2020-03-01
Caldav

aaronoppenheimer:
First functioning caldav feed:

- added caldav.py in lib to hold the caldav-specific stuff
- added helpers to member to manage generating the streams based on member prefs
- added cal_feed_id to the member as a uuid to be used in the calfeed url
- added the cal feed url

So, at this point, member/calfeed/<uuid> generates and returns the caldav feed for that user.

Also added the beginning of the filesystem-based cache - getting that working is next.

Also also did some random bugfixing.


lib/caldav.py
```diff
@@ -0,0 +1,71 @@
...
+        return '{0}:{1} {2}'.format(event.band.name, event.title, event.status)
+
+    def _make_description(event):
+        """ description is the details, plus the setlist """
+        x = ''
+        if e.details:
+            x = '{0}'.format(e.details.replace('\r\n', '\\n'))
+        if e.setlist:
+            x = '{0}\\n\\n{1}'.format(x, e.setlist.replace('\r\n', '\\n'))
+        return x
```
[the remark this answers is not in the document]
aaronoppenheimer: Hm - actually, I just lifted this from the go2 code - will check to see if we need it.

aaronoppenheimer:
OK!

I was sloppy with formatting; the IDE I'm using doesn't give me a hard time like my old one. I did some automatic pep8 which may be more than I want, but I'm happy to have the ugliness pointed out.

I split the tests up per your recommendation.

At this point, it's all working...except that I can't test the automatic updating because django-q has problems on the mac. I can't run 'manage.py qcluster' due to some error in the bowels of multiprocessing. I can make qcluster run synchronous but that doesn't seem to run the scheduled tasks. 

So: If you can a chance, can you test it?

- create a 'calfeeds' directory at the 'manage.py' level
- start the server and qcluster
- create a gig
- in the admin panel, schedule a task to call member.helpers.update_all_calfeeds
- when it runs, you should find a file in calfeeds named after your user's cal_feed_uuid.txt
- if you access /member/calfeed/[uuid] you should get a reasonable ical-formatted file out of it

rschroll:
After a couple of false starts, I got it to work. I've pushed some fixes up here.  There are two other issues that I noticed:
1. `make_calfeed` fails on events that don't have an `enddate` set.  I don't know what the correct thing to do here is.  Can an event not have an *dtend*?  If so, what do clients display?  We may need to make up an end time for this purpose.
2. A member changing their calendaring preferences should mark their feed as dirty.  Similarly, a member confirming attendance should mark their feed as dirty, at least if they've chosen to only get confirmed events on their feed.

The former we probably want to fix now, but the latter could become issues to be solved later.

rschroll:
From the RFC, it looks like *dtend* is not required, but a duration of 1 day is assumed when it's missing.  This is probably not what we want.

lib/caldav.py
```diff
@@ -0,0 +1,82 @@
...
+        cal.add('prodid', '-//Gig-o-Matic//gig-o-matic.com//')
+        cal.add('version', '2.0')
+        cal.add('X-WR-CALNAME', the_title)
+        cal.add('X-WR-CALDESC',
+                '{0} {1}'.format(_('Gig-o-Matic calendar for'), the_title))
+        for e in the_events:
+            with timezone.override(e.band.timezone):
+                event = Event()
+                event.add('dtstamp', timezone.now())
+                event.add('uid', the_uid)
```
rschroll: Presumably, this should be a UID associated with the event, rather than with the whole calendar, right?

lib/caldav.py
```diff
@@ -0,0 +1,82 @@
...
+                event.add('dtstart', e.date)
+                event.add('dtend', e.enddate)
+                event.add('description', _make_description(e))
+                event.add('location', e.address)
+                event.add(
+                    'url', 'http://www.gig-o-matic.com/gig/{0}'.format(e.id))
+                # todo don't hardwire the URL
+                # todo go2 also has sequence:0, status:confirmed, and transp:opaque attributes - need those?
+                cal.add_component(event)
+    return cal.to_ical().decode('ascii')
```
rschroll: It appears that ical files use UTF-8 encoding (https://tools.ietf.org/html/rfc5545#section-3.1.4).  That said, since we're going to be putting this on disk and sending it over the wire, I wonder if we want to leave this as bytes.

lib/caldav.py
aaronoppenheimer: Yes - will fix.

lib/caldav.py
aaronoppenheimer: Probably so. This is something I mangled badly in go2. Will fix!

aaronoppenheimer:
Thanks a ton. For gigs without end dates, I'll just put in an hour so it shows on the calendar in the right place - that's how go2 works today.

Issue #44, opened 2020-03-08
Failing timezone test


aaronoppenheimer:
https://github.com/SecondLiners/GO3/blob/11952ae0a1f47eda4c738d4f34c8b50b648bf433/gig/tests.py#L173

This is failing because:
- band timezone is set to New York on line 166
- gig is created for noon in band's timezone
- email is checked to see if time is properly offset for timezone - but test expects time to be shifted as if the gig date was in UTC

What's the intent? Should we be leaving the band tz in default UTC?

rschroll:
This test was created in 30da2e1, and I wasn't quite sure what it's trying to test.  From the test name, it seems to be testing that the times are localized correctly.  But this is implicitly tested for english in `test_new_gig_email` and explicitly tested for german in `test_new_gig_localization`.  Since it was also adjusting time zones, it might also be testing that we're storing and interpreting time-zone-aware dates correctly.

Originally, both the band and the gig were in UTC, but we switched to ET to prepare the email.  This was before we set the timezone of the email from the band.  With this change, I set the band to ET, while setting the gig time in UTC, to the same result.  However, commit 6650dcf sets the test gig TZ from the band TZ, making this split impossible.  After the merge, this test fails.

So, if the purpose was to test localization of times, I think it's already covered and we could get rid of this test.  If the purpose was to check that times are correctly displayed when there's a mismatch between gig and band timezones, then we need to ensure that these two are different.

aaronoppenheimer:
Well, no idea what happened, but no matter.

Gig and band timezones will almost always be the same. Sometimes a band might travel to another zone to play a gig at Honk!Timbuktu or something, so in that case it'd be nice to let them override the TZ for a gig so that things show up properly on calendars. So I'll update this to explicity demonstrate that a gig with one TZ shows up properly when the band has another - it's probably redundant but since timezones are a pain, more test is better.

Thanks for digging into this!

A

aaronoppenheimer:
Hey, you already fixed it! Thanks.

Issue #37, opened 2020-02-29
Time and date information not well formatted in gig emails


rschroll:
I see two issues right now:
1. If there is both a `date` and `enddate` on a gig, the email will display "Date: start - end", even if those are the same dates.
2. If the `date`, `setdate`, and `enddate` are on different days, the email will display a date range and the three different times, but not say on which day each time is.  Most of the time it will be obvious, but I can imagine times when it's not.

rschroll:
My proposal for dealing with this:
1. If the dates are all within 24 hours of each other [1], display the date of the start date in a Date line and all available times in a Time line.  This is almost as it is now.
2. If not, display each available date on its own line, with both the date and time displayed.  This will look different, but in this case will probably be more readable.

Thoughts?

[1] I originally was thinking "on the same day (in the band's timezone)", but I could imagine a gig with a set date of 7/1 11PM and an end date of 7/2 1AM.  In most people's minds, this would be a one-day gig on 7/1, so that's what we should display.  When the end time is listed as 1 AM, people should be able to figure out that it's the 1 AM after the 11 PM, not before.

aaronoppenheimer:
Yes, agreed this makes sense.

A

PR #45, opened 2020-03-08
Improve display of gig times in gig emails

rschroll:
This follows the discussion in #37.  I use `SHORT_DATETIME_FORMAT` to display the date and time together, but in English this produces strings of the form "01/02/2100 noon".  I think it would look better the other way around, but I'm not sure the best way to do that.  I think we can redefine `SHORT_DATETIME_FORMAT`, but I don't know if we can do that for all of the translations.  We could also just pass our own custom format in (which would allow us to include the day in a single formatting), but I don't know how translations work there either.  I'll investigate, but if you know about such things, I'm happy to hear.

These changes made fix 2 for #44 relatively easy, although I'm still not sure that's what we mean to be testing.

### DIFF
--- a/gig/helpers.py
+++ b/gig/helpers.py
@@ -103,6 +103,12 @@ def generate_changes(latest, previous):
         changes.append((_('Details'), _('(See below.)'), None))
     return changes
 
+def is_single_day(gig):
+    end = gig.enddate or gig.setdate
+    if not end:
+        return True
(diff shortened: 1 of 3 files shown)


aaronoppenheimer:
Apparently, if you use L10N, you can override the formats pretty easily per locale:

https://stackoverflow.com/questions/3617804/django-use-of-date-format-datetime-format-time-format-in-settings-py

PR #49, opened 2020-04-05
Invite machinery

rschroll:
This provides most of the machinery to allow invites.  Unlike the previous code, this introduces a new `Invite` object in the database, which keeps track of the invite email, band, and language.  Only when a user accepts an invite do we try to associate it with a Member, either existing or creating a new one.  This helps us avoid all sorts of odd conditions with members being created after invites.

When a user visits the invite accept page, the invitation is automatically accepted if they are logged in as the member in question or are not logged in at all.  (In the latter case, we assume they've logged out at some point, so we make this work like the plan status updates -- visiting is enough to cause an action.)  If they are logged in as another user, we give them the option to steal the invite for their current member, or log out to accept it at the original email address.

If the user is trying to accept an invite for an email address that doesn't have a Member, they are prompted to create a new member.  After member creation, they are logged in and redirected back into the invite accept workflow, which ensures the invite gets accepted.  I realized that we could (ab)use this workflow to allow new members to sign up without a band invite.  I created the possibility for new users to trigger an Invite without an associated band.  This will send them an invite email, ensuring that they control the email address in question, and then then invite accept flow will push them through the Member-creation workflow.  I don't know if this is a great idea or not; we could snip it out easily without breaking the rest of the machinery.

Some things that could use some more thought, either now or later:
- There is nothing currently to trigger the invite creation endpoints, and these endpoints just return JSON payloads.  The GO2 code used AJAX to create the invites, and I didn't know if we wanted to keep that in place or fall back to HTML form submission.  Even if we keep this for band invites, the sign-up invites probably need a form submission solution.  And if they are to be AJAX, code needs to be written to trigger them.
- There are guards on the various processing endpoints to avoid creating multiple Members or Assocs.  This means it can survive having multiple invites for the same Member/Band pair.  But should we allow this to happen?  We could prevent creating invites that would be duplicates.  We could delete old invites that match the settings of new invites.  But I worry that both of these could cause confusion for a user that's having trouble finding invite emails and is asking the band admin to resend them.  Leaving all invites valid increases the chance the one he eventually finds actually works.  If we're worried about outstanding invites, we could try deleting matching invites as soon as one is accepted.
- When an invite is accepted, it is currently deleted from the database.  Would it be better to keep it around but mark it as "accepted"?  This could block processing of it, while allowing us to see all the past invites for debugging purposes.  It would also let us give different error messages when you visit an invite page that has already been accepted compared to an invite page for an invite that doesn't exist.  The downside is keeping a lot of stuff in the database that we really don't need.
- I tried to lay out the new pages I created following the model of existing pages.  It may be worth coming up with some design guides.  Or maybe we want to wait until most things are ready and just do a design pass on everything at once.


member/models.py
```diff
@@ -164,3 +165,23 @@ class MemberPreferences(models.Model):
...
+    language = models.CharField(choices=LANGUAGES, max_length=200, default='en')
+
+    @property
+    def email_line(self):
+        return self.email
+
+    # We want language to work as it does for members
+    @property
+    def preferences(self):
+        return SimpleNamespace(language=self.language)
```
aaronoppenheimer, line 187: I don't understand - what's the thinking with this?

aaronoppenheimer:
Hi Robert - this is great, sorry it has taken me so long to get around to it.

re: your comments - I'm happy to move to an HTML-based version of the invitation process. I can't remember why I bothered doing it as AJAX last time.

I think we needn't bother keeping invitations around, but maybe we should add a "date created" and "date confirmed" to the assoc itself. If it's through an invite, the "date created" and "date confirmed" are the same, but if I ask to join a band the assoc is created; it's confirmed by a band admin later. That way we know when someone joined without keeping the invites - we can just delete all invites that are accepted, and can delete invites that have a matching email/band when any other invite is accepted. Also - sometimes I've seen an invite go to someone and then they just go right to the gig-o and create an account. If they do that, should we just create accept any outstanding invites for them? Or, if they ask to join a band, should we delete any invites that would have had them join that band had the accepted it?

Re: the layouts - I'm hip-deep in figuring out how to best work with the django form system; a style-guide is a good idea but at this point I'm not sure things won't change. I go back-and-forth on a visual refresh, too. So what you've done is great for now.

member/models.py
rschroll: I wanted to be able to reuse member.helpers.prepare_email with Invite objects, since that takes care of translation and HTML formatting.  This function is expecting Member objects, but the only part of that API I was using was the `.email_line` and `.preferences.language` attributes.  This code duplicates those attributes in the Invite object, so it can be used here as well.

With some distance, I don't particularly like this solution.  Here's two other ideas:
1. Add a `.language` property to Members, which looks up the language in their preferences.  Then, `prepare_email` can use that API, and we don't have to jump through hoops here.
2. Create a `EmailRecipient` class / dictionary / named tuple that `prepare_email` will accept, and then build a way for both Member and Invite objects to create such objects.  This would keep us from accidentally using properties of one that the other doesn't have.

Any thoughts?

rschroll:
I went ahead with option 2: A separate class to pass recipient information into the email system.  This is arguably overkill, since I don't anticipate more types of email recipients than what we already have.  But it does help isolate the email system from the member classes a bit, which is probably a good thing.  It also gave me a chance to use dataclasses for the first time, which is nice.

As for outstanding invites, I had been thinking of adding an "Invites" section to the user's profile page.  Since that's where we're currently redirecting people after making an account, they'd get a chance to accept those straight away.  I think that's a better approach than just auto-accepting everything.  I can start working on the forms for creating invites, but it may be better to handle them in a separate PR.  Otherwise, I fear this one is getting pretty large.

PR #50, opened 2020-04-05
Set up automated pylint testing

rschroll:
This happened a bit accidentally.  I don't know if we want it or not, but we might as well take a look.

My editor (VSCode) automatically highlights pylint errors.  This is occasionally useful, but I was getting a lot of false positives because pylint doesn't understand the tricks the Django ORM is pulling.  But it turns out that there's a pylint-django package that fixes that.  I installed it, but you also have to tell pylint to use it.

I could have done that via an editor config, but I thought that anyone else using pylint would want this too, so I put it in a pylintrc file.  However, this triggered VSCode to switch from using it's list of rules to using all rules that pylint has.  And there are a lot of those!  So I copied VSCode's list of rules into the pylintrc file.  There may be some we don't care about, and there may be others that we do, but this seemed like a more reasonable starting place that "everything".

Of course, if we're keeping in the repo the list of pylint rules we care about, it's easy to test those in CI.  So I set that up too.

You'll notice that tests are not passing, because pylint complains about problems in our code.  The most common issue is unused variables, and most of these seem to be valid complaints.  Most of the rest seem to be false positives, so we'd have to disable the rule, either for the line or globally.  Don't know if we want to go down this route or not.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -14,6 +14,7 @@ jobs:
       with:
         python-version: 3.8
     - name: Install dependencies
+      id: pip
       run: |
         python -m pip install --upgrade pip
         pip install -r requirements.txt
@@ -22,4 +23,5 @@ jobs:
         python manage.py collectstatic
         python manage.py test
     - name: Check with Pylint
+      if: ${{ always() && steps.pip.outcome == 'Success' }}
       run: modules=(*/__init__.py) && pylint -j0 ${modules[@]%/*}
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -21,3 +21,5 @@ jobs:
       run: |
         python manage.py collectstatic
         python manage.py test
+    - name: Check with Pylint
+      run: modules=(*/__init__.py) && pylint -j0 ${modules[@]%/*}
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,7 @@
+[MASTER]
+load-plugins=pylint_django
+
+[MESSAGES CONTROL]
+disable=all
+
+enable=F,E,unreachable,duplicate-key,unnecessary-semicolon,global-variable-not-assigned,unused-variable,binary-op-exception,bad-format-string,anomalous-backslash-in-string,bad-open-mode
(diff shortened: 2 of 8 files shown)


aaronoppenheimer:
I think this is a fine idea (and sorry it took me so long to get to this). It'll be a pain to bring the code up to snuff but worth it, I think. Still think it's a good idea?

rschroll:
I've cleaned up the simpler problems.  Here's what remains:

```
gig/models.py:29:30: E0602: Undefined variable 'section' (undefined-variable)
```
I think this should be a string.
```
gig/models.py:105:20: E0602: Undefined variable 'StatusOptions' (undefined-variable)
gig/models.py:109:20: E0602: Undefined variable 'StatusOptions' (undefined-variable)
```
Technically a false positive, I think, but it would be clearer as `AbstractGig.StatusOptions`.  But but, in other places we have `IntegerChoices` as separate classes in a `utils` module.  Should we do that here?  But but but, do we want `=` or `==` here?
```
gig/models.py:150:15: E1101: Instance of 'AbstractGig' has no 'plans' member (no-member)
```
An actual problem -- plans only exist on the concrete `Gig` class.  Should we move this whole property over?
```
gig/tests.py:410:49: E1101: Instance of 'tuple' has no 'label' member (no-member)
gig/tests.py:422:49: E1101: Instance of 'tuple' has no 'label' member (no-member)
```
It looks like it doesn't know that Django is doing inside of these `IntegerChoices`.  Probably just have to ignore these.
```
member/models.py:136:39: E1101: Instance of 'Member' has no 'preferences' member (no-member)
```
Looks like it isn't picking up on the `OneToOneField` with `MemberPreferences`.  It's a bit odd because it does get the `ForeignKey` relationships.
```
member/views.py:53:8: W0612: Unused variable 'ok_to_show' (unused-variable)
```
I'm not sure what's actually going on in this function....
```
member/tests.py:745:8: E1101: Instance of 'InviteTest' has no 'assertRenderedWith' member (no-member)
member/tests.py:756:8: E1101: Instance of 'InviteTest' has no 'assertRenderedWith' member (no-member)
```
A real problem, and the reason the tests are failing.  (See #61).

PR #51, opened 2020-05-03
Add UI around invite machinery

rschroll:
This includes:
- An invite page for a band to invite new members.
- A list of pending invites on the band detail page, with the ability to delete them.
- A list of invites on the member detail page, with links to accept or decline them.

All of this is done with straight server-side processing, despite the previous version and, in some cases, neighboring code, using various AJAX approaches.  I'm happy to make it more AJAXy if desired, but this should work okay.

### DIFF
--- a/member/views.py
+++ b/member/views.py
@@ -217,11 +217,13 @@ def accept_invite(request, pk):
         if invite.band and Assoc.objects.filter(band=invite.band, member=member).count() == 0:
             Assoc.objects.create(band=invite.band, member=member,
                                  status=AssocStatusChoices.CONFIRMED)
+            if request.user.is_authenticated:
+                # We'll be redirecting them to their profile page, and we want to display:
+                messages.success(request,
+                                 format_lazy(_('You are now a member of {band}.'),
+                                             band=invite.band.name))
         invite.delete()
         if request.user.is_authenticated:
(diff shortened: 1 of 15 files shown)


member/views.py
```diff
@@ -143,34 +151,57 @@ def get_context_data(self, **kwargs):
...
+        if in_band:
+            messages.info(self.request,
+                          format_lazy(_('These users are already in the band: {e}'),
+                                      e=join_trans(_(', '), in_band)))
+
+        return super().form_valid(form)
 
-    return JsonResponse({'invited': invited, 'in_band': in_band, 'invalid': invalid})
+    def get_success_url(self):
+        return reverse('member-invite', args=[self.kwargs['bk']])
```
aaronoppenheimer, line 204: If all of the emails are valid, we should return to the band page, not the invite page

rschroll:
Good catch.  I wasn't 100% sure what to do if there were already-joined emails in the list submitted.  I decided to treat that like an error and redirect back to the original form, in case this was a sign that the user entered the wrong list.  But it's a simple change to treat this as a success condition and go to the band details page.

aaronoppenheimer:
Yeah, I think if you invite someone already in the band we should just declare victory and move on. Thanks, Robert!

PR #56, opened 2020-05-10
Help pages

rschroll:
They're mostly static, but the is a form submission for the band request page.  We'll want to check that this info is up-to-date; some of it is definitely not!

I've built this on top of the auth-templates branch (see #54).  I've made this PR against that branch, so we can more easily see what's changed.  Either we can merge this into that, and then into master, or we can merge that one first and retarget this onto master.


aaronoppenheimer:
Hi Robert - I've merged in #54 so if you want to rebase this onto the master we can just pull it in directly. I don't know how to do that...

rschroll:
I think we just have to change the base branch, as I just did.  It should be able to merge into master, but if not, let me know and I'll rebase.

aaronoppenheimer:
How do you change the base branch?

PR #62, opened 2020-05-11
Datepicker2

aaronoppenheimer:
Transitioned to a more modern datepicker for git edits, figured out how to get it to work with translations, timezones, bootstrap. The gig_form template is still a bit of a mess and there's more logic to add to make sure the dates are valid, but things are at least stable. I had to learn way more than I wanted to know about form handling, and I had to restart the whole project about five times but FIFTH TIME IS THE CHARM.


aaronoppenheimer:
Also fixed the outstanding pylint issues

gig/models.py
```diff
@@ -147,7 +135,7 @@ def member_plans(self):
         for a in absent:
             Plan.objects.create(gig=self, assoc=a, section=s)
         # now that we have one for every member, return the list
-        return self.plans
+        return getattr(self, 'plans') # technically the gig object has no 'plans' attribute; it's defined by the ORM system
```
rschroll, line 138: I think the pylint error here was not due to the fact that it couldn't track the foreign key, but because it's Gig, and not AbstractGig, that Plan maps to.  At the moment, this will never be a problem.  But if we get another descendant of AbstractGig, this could be a bug.

member/models.py
```diff
@@ -133,7 +134,7 @@ def motd(self):
 
     def as_email_recipient(self):
         return EmailRecipient(name=self.username, email=self.email,
-                              language=self.preferences.language)
+                              language=getattr(self,'preferences').language)
```
rschroll, line 137: You can turn off a pylint error for a single line by adding a comment: `# pylint: disable=error-name`.  I'd prefer that to this sort of trickery.

rschroll: And it looks like you already knew that.  But I still think it's appropriate here as well.

rschroll:
A couple comments on the pylint fixes.  Let me know if and when you'd like a more careful review.

aaronoppenheimer:
Hi Robert - I don't understand. The 'plans' object has a foreign key that points to 'Gig' with the reverse-reference to plans. But if I change it to point to AbstractGig, pylint gives me the same error. It's right, there's no real 'plans' attribute on AbstractGig, it's faked by the ORM system, isn't it?

A

On Thu, May 14, 2020, at 3:14 AM, Robert Schroll wrote:
> 

> *@rschroll* commented on this pull request.

> 
> In gig/models.py <https://github.com/SecondLiners/GO3/pull/62#discussion_r424918318>:

> > @@ -147,7 +135,7 @@ def member_plans(self):
         for a in absent:
             Plan.objects.create(gig=self, assoc=a, section=s)
         # now that we have one for every member, return the list
-        return self.plans
+        return getattr(self, 'plans') # technically the gig object has no 'plans' attribute; it's defined by the ORM system

aaronoppenheimer:
Oh - when I try to migrate to using the abstract model I get an error. I apparently need to use a GenericForeignKey. Will figure it out.

A

On Thu, May 14, 2020, at 6:24 PM, Aaron Oppenheimer wrote:
> 
> Hi Robert - I don't understand. The 'plans' object has a foreign key that points to 'Gig' with the reverse-reference to plans. But if I change it to point to AbstractGig, pylint gives me the same error. It's right, there's no real 'plans' attribute on AbstractGig, it's faked by the ORM system, isn't it?
> 
> A
> 
> On Thu, May 14, 2020, at 3:14 AM, Robert Schroll wrote:
>> 

>> *@rschroll* commented on this pull request.

>> 
>> In gig/models.py <https://github.com/SecondLiners/GO3/pull/62#discussion_r424918318>:

>> > @@ -147,7 +135,7 @@ def member_plans(self):
         for a in absent:
             Plan.objects.create(gig=self, assoc=a, section=s)
         # now that we have one for every member, return the list
-        return self.plans
+        return getattr(self, 'plans') # technically the gig object has no 'plans' attribute; it's defined by the ORM system

aaronoppenheimer:
Well. One can't use an abstract class as a foreign key if you want to use the 'related name' to do reverse relations - so having Plan point to AbstractGig is a no-go. You can use 'GenericForeignKey' but that doesn't come with a reverse relation so it's not clear how to make it work.

It's definitely wrong for the AbstractGig to have the attribute that refers to 'self.plans' when the plans are defined on the Gig class. So we can move the member_plans attribute to the Gig. This doesn't solve the original problem - pylint still complains because it just doesn't understand the related_name in the foreign key on Plan. But that can be switched off.

The only downside to putting the member_plans attribute on the concrete Gig is that the AbstractGig now has no notion of 'plans' (technically, it never did, I know). There are other types of events I can think of that can be based on the AbstractGig, like a poll, and they have "plans" too. But maybe the better thing to do is make the AbstractGig a concrete class, and package up the things that make a Gig a Gig in a 'details' class which is just referenced by the AbstractGig. Different kinds of gigs then can have plans and all that, but just have different types of 'details' to which they refer. I think that solves all of these problems (except for the pylint thing).

Other thoughts?



On Thu, May 14, 2020, at 6:28 PM, Aaron Oppenheimer wrote:
> 
> Oh - when I try to migrate to using the abstract model I get an error. I apparently need to use a GenericForeignKey. Will figure it out.
> 
> A
> 
> On Thu, May 14, 2020, at 6:24 PM, Aaron Oppenheimer wrote:
>> 
>> Hi Robert - I don't understand. The 'plans' object has a foreign key that points to 'Gig' with the reverse-reference to plans. But if I change it to point to AbstractGig, pylint gives me the same error. It's right, there's no real 'plans' attribute on AbstractGig, it's faked by the ORM system, isn't it?
>> 
>> A
>> 
>> On Thu, May 14, 2020, at 3:14 AM, Robert Schroll wrote:
>>> 

>>> *@rschroll* commented on this pull request.

>>> 
>>> In gig/models.py <https://github.com/SecondLiners/GO3/pull/62#discussion_r424918318>:

>>> > @@ -147,7 +135,7 @@ def member_plans(self):
         for a in absent:
             Plan.objects.create(gig=self, assoc=a, section=s)
         # now that we have one for every member, return the list
-        return self.plans
+        return getattr(self, 'plans') # technically the gig object has no 'plans' attribute; it's defined by the ORM system

rschroll:
It looks like Django supports concrete base classes: https://realpython.com/modeling-polymorphism-django-python/#concrete-base-model.  This seems to do automatically what you were considering with the details class on Gig.  Maybe it's worth a try.  That page warns of performance issues, since new we need a join to pull back all of the details, but I suspect we aren't going to be near database limits.

aaronoppenheimer:
Yes, that can work, though it gave me some other problem and I can't remember what it was now.

In the end, the way "plans" are defined is pretty specific to a Gig anyway. If I wanted to create another kind of event, like a Poll, it would have a different way to represent answers and wouldn't need a lot of the stuff in the Plan class.

When I looked at this in Go2 I just repurposed the Gig as a Poll and ignored all the Gig stuff that don't mean anything to a Poll, like "dress code." But clearly I could abstract out BaseEvent and BaseResponse and derive Gig and GigResponse, Poll and PollResponse. Cleaner but a little more invasive. But I guess the time to do that is now.

(It would still make pylint upset, but whatever.)

aaronoppenheimer:
OK, in the end I think it should stay as it is. I don't really need polymorphism - gigs and polls have some common attributes but beyond that there's no need for real polymophism. And having 'plans' point directly to 'gigs' instead of to the base class is fine. Polls will have an entirely different kind of response. So I submit it as-is (though I changed the name of the abstract base class to AbstractEvent).

PR #63, opened 2020-05-18
Test use of htmx on member detail page

rschroll:
This weekend, I saw the announcement of [htmx](https://htmx.org), a JS library to do live substitution of server-rendered HTML.  Since we do that by hand on some of the pages, I thought it deserved a closer look.  As a test case, I used it to provide the interactivity on the member detail page.  I'll put this here for us to look at and see if this is worth pursuing further.

The member detail page loads in the associations fragment from the server.  This used to be accomplished by some JS that ran on page load.  With htmx, we can just set the `hx-get` attribute on `#bandlist` to indicate where to get the fragment, and the `hx-trigger="load"` attribute to indicate that this should happen on page load.

Our JS had waited for this to succeed, and then it got the other band list.  With this, we can just include an element with `hx-` attributes in the associations fragment, and that too gets retrieved and inserted for us.

We had relied on additional JS to set actions to create an association for each of these other bands.  Here, we can set the `hx-post` attribute on the other band items to cause a POST request be sent to the endpoint to create an assoc.  Before, we'd listen for the response, and reload the associations fragment once it completes.  With the `hx-target` attribute, we can tell htmx to replace the whole fragment with the response to this POST.  To make this work, the server redirects to the assoc list endpoint after handling the request.

Having gone through this, I see some clear advantages:
1. Less code.  This cuts 85 lines, including most of the custom JS.  Fewer lines mean fewer bugs.
2. Clearer behavior.  Before, you would have to keep track of JS from several different templates, and it would not be clear from the HTML alone what was going on.  The `hx-` attributes put the actions right in the HTML.

Not everything is rosy:
1. The library is brand-new, so we don't know if it will last.  I've already found and submitted some bugs and fixes.  It is a re-write of the author's previous project, so we have some confidence.  The library is also relatively small and pure JS, so maintaining it ourselves wouldn't be completely crazy.
2. There's some ugliness around the delete modals.  Maybe there's a cleaner solution here, but I haven't found it.  There's actually a `kt-confirm` attribute, but that would require giving up the Bootstrap modal for the browser dialog.
3. The fact that we look at the response from POST requests to figure out the new content feels like mixing concerns.  The POST ought to update the database, and then the page should figure out how to update itself.  This would cause a problem right now on the band details page, which has the same assoc delete endpoint, but would want a different display.

    We could get this through some custom JS, but that feels like it's undoing the promise of using htmx instead of JS.  I feel like a reasonable feature request would be to add a `hx-` attribute to indicate a redirect after a success of a POST.  The author's been responsive to my bug reports, so if this is our only blocker, I'd be happy to make the case for adding this.

Let me know what you think!


rschroll:
And of course, as soon as I post the above, I stumble across the solution for problem (2).

aaronoppenheimer:
Huh - that is pretty great. It seems like it would make a lot of things easier. I understand the issue with #3 and it seems reasonable to get him to add a 'redirect' tag. #1 - always a risk, right? I don't know at what point we decide "yes, it'll be around for a while." But it's really attractive...

rschroll:
I'll raise the issue of the redirect tag, and see what happens.  I feel like, since this is a pure JS thing, as long as it's working when we adopt it, it's not going to break any time in the near future.  So support is not really a high priority.

aaronoppenheimer:
I agree.

aaronoppenheimer:
I still think this is pretty cool. I see your back-and-forth with the author - satisfied that this will stay useful? If so I'll merge it in.

rschroll:
I think it will work, but I'd like to rework this based on some of the
discussion going on. As it is, we won't be able to have the same API
endpoint cause different reloads. But one of these solutions will fix that.
I'll play with it some more, and I'll let you know when this is ready.

aaronoppenheimer:
OK!

rschroll:
I'm feeling much better with some of the new htmx features.  We don't have to have the API endpoints returning redirects to data.  Also, I've cleaned up the tfparams endpoint.  I'll see about converting the band details page tomorrow, and then this should be good for a merge.

aaronoppenheimer:
OK! When you get a chance, I'd appreciate a walk-through of how the extension stuff works - it's not yet in the htmx documentation (or, I can't find it).

rschroll:
You can find updated documentation on the dev version of the htmx website: https://dev.htmx.org/.  Extensions are here: https://dev.htmx.org/extensions/.  All mine are doing are adding event handlers on specific elements.

BTW, they're planning on a new release of htmx on Monday.  We'll want to upgrade to that, rather than using the random dev version I've linked to right now.

rschroll:
I've moved to the newest release of HTMX and gotten the band detail page over to use it.  The existing structure was a bit odd -- there were several AJAX requests that triggered a page reload when they were finished.  I think this is because they changed the membership in a way that affects several sections of the page.  For now, I've just replicated that behavior.  We can probably do better in the future.

Also, I think I've uncovered a vulnerability -- any user could set themselves to be a band admin, if they can figure out the right URL to hit.  I'm working on a fix now.

aaronoppenheimer:
Are we good to merge this at this point?

rschroll:
Yep! Plenty of additional work to do, but I think the fundamentals
look good.

PR #64, opened 2020-05-22
Added duplicate functionality from the gig detail page. This basicall…

aaronoppenheimer:
…y uses the 'create' path, but populates the edit form from the gig we're duplicating.

Based on 'datepicker2' branch which should be merged first.

### DIFF
--- a/gig/views.py
+++ b/gig/views.py
@@ -20,6 +20,7 @@ from django.views import generic
 from django.urls import reverse
 from django.utils import timezone
 from django import forms
+from django.http import Http404
 from .models import Gig, Plan
 from .forms import GigForm
 from .util import PlanStatusChoices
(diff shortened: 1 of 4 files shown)


aaronoppenheimer:
To clarify, this is "duplicate this gig by pressing the 'duplicate' button" not some functionality that duplicates something else.

gig/views.py
```diff
@@ -114,11 +116,33 @@ def form_valid(self, form):
...
+        # ...but replace the title with a 'copy of'
+        kwargs['initial']['title'] = f'Copy of {kwargs["initial"]["title"]}'
+        return kwargs
+
+    def get_band_from_kwargs(self, **kwargs):
+        # didn't have the band from the request args, so pull it from the gig
+        try:
+            return Gig.objects.get(id=self.kwargs['pk']).band
+        except Gig.DoesNotExist:
+            raise Http404('No gig found matching the query')
```
rschroll, line 140: Perhaps a place to use `get_object_or_404`.

templates/gig/gig_detail.html
```diff
@@ -28,7 +28,7 @@
                                 <a class="btn btn-primary btn-sm" href="/gig/{{ gig.id }}/update">{% trans "Edit" %}</a>
                             {% endif %}
                             {% if user_can_create %}
-                                <a class="btn btn-primary btn-sm" href="gig_edit.html?gk={{ gig.id }}&dupe=1">{% trans "Duplicate" %}</a>
+                                <a class="btn btn-primary btn-sm" href="/gig/{{ gig.id }}/duplicate">{% trans "Duplicate" %}</a>
```
rschroll, line 31: I'd recommend the {% url %} function.

gig/views.py
aaronoppenheimer: "perhaps" - you're too nice

PR #65, opened 2020-05-23
Address handling

aaronoppenheimer:
added simple feature to allow use of URLs in address field without turning them into google maps queries (which I just added to Go2 since 'address' means 'zoom' these days)


aaronoppenheimer:
I found a validation library that applies a big honking regexp to determine if it's a url - but still need a scheme. So I test as-is and with http:// tacked to the front.

PR #66, opened 2020-05-23
Added handling of gig comments

aaronoppenheimer:
Based this on top of **htmx** which made it super easy.


gig/urls.py
```diff
@@ -31,4 +31,7 @@
     path('plan/<uuid:pk>/section/<int:val>', helpers.update_plan_section, name='plan-update-section'),
 
     path('answer/<uuid:pk>/<int:val>', views.answer, name='gig-answer'),
+
+    path('<int:pk>/comments', views.CommentsView.as_view(), name='gig-comments'),
+
```
rschroll, line 36: Nit: I'd sort this up with the other views for the gig.

gig/views.py
```diff
@@ -17,15 +17,16 @@
...
 from django import forms
-from .models import Gig, Plan
+from .models import Gig, Plan, GigComment
 from .forms import GigForm
 from band.models import Band
 from gig.helpers import notify_new_gig
-
-
+from django.contrib.auth.mixins import LoginRequiredMixin
+import urllib.parse
```
rschroll, line 29: Nit: Keep blank lines between imports and code.

gig/views.py
```diff
@@ -101,6 +102,29 @@ def form_valid(self, form):
...
+        gig = Gig.objects.get(id=self.kwargs['pk'])
+        context['gig'] = gig
+        try:
+            context['comments'] = GigComment.objects.filter(gig__id=self.kwargs['pk']).order_by('created_date')
+        except GigComment.DoesNotExist:
+            context['comments'] = None
+        return context
+
+    def post(self, request, **kwargs):
+        text = request.body.decode('ascii')
```
rschroll, line 119: Would using `request.POST` let us access this information without this manual processing?

templates/gig/gig_comments.html
```diff
@@ -0,0 +1,32 @@
+{% load i18n %}
+{% load humanize %}
+
+<div class="row">
+    <div class="col-12" id="gig_comment">
+        {% if comments %}
+            {% for the_comment in comments %}
+                <div class='comment-intro'>
+                    {% if the_comment.member %}
+                        <a href="/member/{{the_comment.member.id}}">{{ the_comment.member.display_name }}</a>
```
rschroll, line 10: Use the {% url %} function.

templates/gig/gig_comments.html
```diff
@@ -0,0 +1,32 @@
...
+        {% endif %}
+    </div>
+</div>
+<div class="row mt-4">
+    <div class="col-md-10 col-12">
+        <textarea class="form-control" id="commentinput" name='commenttext' placeholder="{% trans "say your piece!" %}" name="comment_input" rows="2"></textarea>
+    </div>
+    <div class="mx-auto">
+        <button class="btn btn-primary btn-sm" 
+        hx-post="/gig/{{ gig.id }}/comments" hx-target="#gig_commentcard"
```
rschroll, line 29: The {% url %} function can be used here, too.

templates/gig/gig_comments.html
```diff
@@ -0,0 +1,32 @@
...
+        hx-include="[name='commenttext']">{% trans "Say it!" %}</button>
```
rschroll, line 30: Don't know if it's be easier or not, but if we used a form element, htmx would automatically include the inputs within it.  (I think.)

templates/gig/gig_detail.html
```diff
@@ -256,18 +256,20 @@
                     {% trans "Comments" %}
                 </div></div>
             </div>
-            <div class="card-body">
+            <div class="card-body" id="gig_commentcard" hx-get="/gig/{{ gig.id }}/comments" hx-trigger="load">
                 <div class="row">
```
rschroll, line 260: All of the children are being replaced on load, so I think we can just get rid of them.

templates/gig/gig_detail.html
```diff
@@ -324,7 +325,7 @@
 {% endif %}
 
 $(document).ready(function() {
-    init_plan_buttons("{{csrf_token}}")
+    // init_plan_buttons("{{csrf_token}}")
```
rschroll, line 328: I'd just delete these, rather than commenting them out.  If we need them again, they're in the git history.  Otherwise, we're going to wonder if these were important when we notice the comments six months from now.

gig/views.py
aaronoppenheimer: yes!

templates/gig/gig_comments.html
aaronoppenheimer: It does! I don't know if it's easier, either, and maybe hides some detail, but I'm ok with it.

templates/gig/gig_detail.html
aaronoppenheimer: Yes - the javascript has always been messy and now that we're using django to handle some of the form validation, and now htmx, it all needs going over with a fine-tooth comb.

templates/gig/gig_detail.html
aaronoppenheimer: Yes - this is leftover and I forgot to remove.

aaronoppenheimer:
rebased to updated htmx branch

gig/views.py
```diff
@@ -114,10 +116,40 @@ def form_valid(self, form):
...
+        context['gig'] = gig
+        try:
+            context['comments'] = GigComment.objects.filter(gig__id=self.kwargs['pk']).order_by('created_date')
+        except GigComment.DoesNotExist:
+            context['comments'] = None
+        return context
+
+    def get(self, request, *args, **kwargs):
+        if not has_comment_permission(self.request.user, get_object_or_404(Gig, id=kwargs['pk'])):
+            raise PermissionError('Illegal access to comments')
```
rschroll, line 134: Rather than raising an error here, would it be better to return a 401 or a 403?  (I can never remember which is appropriate.)

gig/views.py
```diff
@@ -114,10 +116,40 @@ def form_valid(self, form):
...
+
+        text = request.POST.get('commenttext','').strip()
+        if text:
+            GigComment.objects.create(text=text, member=request.user, gig=Gig.objects.get(id=kwargs['pk']))
+        context = self.get_context_data(**kwargs)
+        return self.render_to_response(context)
+
+
+def has_comment_permission(user, gig):
+    return Assoc.objects.filter(member = user, band=gig.band).count() == 1
```
rschroll, line 149: To make things easier, perhaps this function can trigger a 401 or 403 (again, can't remember), and then the calling functions don't need to worry about the user being unauthorized.

gig/views.py
aaronoppenheimer: agreed, bad habit

gig/views.py
aaronoppenheimer: agreed, bad habit

PR #74, opened 2020-06-08
Changes to the plan update functionality

rschroll:
First, the plan update endpoints are now protected by a decorator limiting access.  This is probably strictly unnecessary, as the random plan id acts as a password, so the URL is hard to guess.  But adding defense in depth seems like a good idea in general.

Second, switch (most of) the client-side update functionality over to HTMX.  The gig comment is left in custom JS, but it doesn't seem to be functioning correctly right now.  I'll update this PR if the fix looks simple, or make another if it gets more involved.


templates/agenda/agenda_plan_edit.html
```diff
@@ -59,17 +59,21 @@
...
+{% endcomment %}
                             {% endif %}
                         </button>
                         <div class="dropdown-menu" aria-labelledby="sel-{{plan.id}}">
                             {% for section in band.sections.all %}
-                                <a class="dropdown-item" onclick="section_select('gig/plan','{{ plan.id }}','sel-{{plan.id}}','{{ section.id }}', 
-                                                                                '{{ section.name }}','{{ csrf_token}}')">{{ section.name }}</a>
+                                <a class="dropdown-item"
+                                   hx-get="{% url 'plan-update-section' pk=plan.id val=section.id %}"
+                                   hx-ext="update-dropdown"
```
aaronoppenheimer, line 69: Why do we do this here, but not on the similar construct in plan_icon_button.html?

rschroll: Because I was translating the existing code pretty literally.  The plan icon button had a function that determined the new HTML to put in based on the status value.  I moved that server side (which was easy, since we already had a template for exactly that) and returned the new value.  The others subbed in the value from that selected dropdown, so I used this "update-dropdown" extension to do that.

Of course, now that you point it out, we could also use that for the plan icon button.  In fact, I think it's better, both for consistency and to separate the front-end behavior from back-end logic.  I'll go ahead and get that in soon.

aaronoppenheimer: Sure, blame whoever wrote the original code. That guy!

Thanks Robert - I'm so glad you found the htmx library. Makes things so much cleaner.

aaronoppenheimer:
Oops - force-merged in the short-agenda-switch code; forgot it was based on this instead of master. It uses htmx to paginate the agenda page.

templates/agenda/agenda_plan_edit.html
aaronoppenheimer: Well, I tried to use the same "update the indicator content from the source-of-event content" approach, but since there's also a label in the menu item, that shows up in the indicator to, which we don't want. So we can leave it how it is for now.

PR #75, opened 2020-06-09
Short agenda switch

aaronoppenheimer:
Updated the agenda page to fetch the list of gigs using htmx; added a "more..." button for long lists instead of short/long list member preference.


aaronoppenheimer:
Oh - there's a whole paginator class so I don't have to do this by hand. Duh.

PR #80, opened 2020-07-31
Cal view

aaronoppenheimer:
Calendar view of agenda


aaronoppenheimer:
This adds the 'fullcalendar' library, which is why there are 100+ files.

PR #211, opened 2024-02-25
Dockerize

bklang:
This PR starts Dockerizing the Gig-o-Matic 3 application. It has two goals:

* Make it easier for new developers to get up and running
* Make it easier to ultimately deploy the to hosting somewhere

This is a draft PR to track the work in progress

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+FROM python:3.11-alpine
+
+WORKDIR /usr/src/app
+
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
+
+RUN apk add --no-cache tini gcc musl-dev mysql-dev
+
+RUN pip install --upgrade pip
+COPY requirements.txt ./
+RUN pip install --no-cache-dir -r requirements.txt
+
+COPY requirements.mysql.txt ./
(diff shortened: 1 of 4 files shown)


jeffwecan:
FWIW, assuming https://github.com/Gig-o-Matic/GO3/pull/220 is included in one form or another, I have an example of utilizing poetry for dependency installation within a multi-stage docker build over yonder: https://github.com/lamurgadeaustin/go3/blob/gcp_main/Dockerfile

That said, there is unfortunately not a universally agreed upon best practice that I found for handling that use case. So I imagine you may find some more preferable way of navigating it yourself 😅.

bklang:
> FWIW, assuming #220 is included in one form or another, I have an example of utilizing poetry for dependency installation within a multi-stage docker build over yonder: https://github.com/lamurgadeaustin/go3/blob/gcp_main/Dockerfile
> 
> That said, there is unfortunately not a universally agreed upon best practice that I found for handling that use case. So I imagine you may find some more preferable way of navigating it yourself 😅.

Great! That is really helpful, thank you.

I'm happy with this PR since it is the basis of the future-production deployment, and I'm using it for local development. I'm going to get this merged then take a look at #220 and the link you referenced above.

Still TODO (but after this PR merges because I'm trying to get launched) is updating developer docs to explain how to use Docker to get up and running from scratch.

PR #217, opened 2024-02-25
Fix failing tests

bklang:
These tests were written against hard coded dates, which are now in the past. Make date calculations relative to current date so tests continue to pass in the future.


aaronoppenheimer (approved):
That's so funny. 2023 was SO FAR in the future, too.

PR #220, opened 2024-02-27
Add Poetry for Dependency Management, Etc.

jeffwecan:
I figure it would be handy to have some process around how we handle dependency management / version resolution within the project and @bklang seemed to be aligned with that sentiment. So this PR adds [poetry](python-poetry.org) to that end. 

To be frank, most of my experience is around basic 'ole [pip-tools](https://github.com/jazzband/pip-tools) for this purpose and I'm relatively new to using poetry. So I won't claim that its the best option out there (python has perhaps too many options nowadays :P). However, I've had a pleasant enough experience with `poetry` in the couple of projects I have used it for.

As far as the actual [version constraints](https://python-poetry.org/docs/dependency-specification/) listed in the pyproject.toml config file go, I (mostly) kept as strict as they had been in the static requirements.txt status quo. That is, they're mostly pinned to specific versions. As opposed to looser constraints that would allow for ongoing minor/patch version dependency updates. In a sense, this is more strict than needed since poetry brings along a poetry.lock file to handle keeping a list of the currently selected versions. We could loosen up those constraints in this PR, but I figured it would be nice to minimize the number of changes made in this initial poetry PR. Subsequently the loosening of those constraints could be navigated as part of other work down the line (e.g., when we have cause to update a dependencies versions, we then loosen the constraint at that point).

As an aside, this PR also includes some not-strictly-related README updates that I'm fine removing upon request.


bklang:
This is really good work. I know I requested it and then let it sit here. To share with you why:
* I've been more focused on fixing functional issues to make sure we can migrate bands before GO2 disappears
* I'm not fully confident I understand how Poetry and Pip work together. It looks like this may be better for local development and for managing upgrades in a more sane manner, but it ultimately depends on Pip for production. So maybe we don't need to touch the Dockerfile at all?

The above is a bit different from my familiarity with how Bundler works in Ruby-land, as Bundler enforces loading of specific library versions. Python just seems to load whatever is installed. So until I have the time to better understand what Poetry does and how it does it, I'm not confident merging it.

My plan is to look at this closer the next time I need to upgrade dependencies.

bklang:
Closing this for now. Thanks again for the contribution, we may revisit this in the future

PR #221, opened 2024-02-27
Add `makemigrations --check` Step

jeffwecan:
To ensure we're always including requisite migrations with any associated model/etc. changes.

(Leaving this as a draft PR to allow https://github.com/Gig-o-Matic/GO3/pull/220 to land first)


.github/workflows/testing.yml
```diff
@@ -29,5 +29,5 @@ jobs:
           poetry run python manage.py collectstatic
           poetry run python manage.py test
 
-      - name: Check with Pylint
-        run: modules=(*/__init__.py) && poetry run pylint -j0 --load-plugins=pylint_django --django-settings-module=go3.settings ${modules[@]%/*}
```
bklang: Is it your recommendation to stop using pylint, or was this accidental?

jeffwecan: That was inadvertent and this PR can just stay closed until https://github.com/Gig-o-Matic/GO3/pull/220 is reviewed for clarity's sake.... 😄

PR #231, opened 2024-02-29
Migration: import band data from CSV

bklang:
This is a quick and dirty tool to import migrated band data from GO2. To be deleted once all bands are migrated.


bklang:
It's quick and dirty, but it works now.

PR #243, opened 2024-03-02
Change default new gig status back to Unconfirmed

bklang:
Feedback from users is that "Unknown" sounds like an error. GO2 used "Unconfirmed" which seems more descriptive to me.


bklang:
@aaronoppenheimer Pinging you on this one in case you have strong feelings about it?

aaronoppenheimer:
Unconfirmed is better. I can't remember changing it to "unknown" for any particular reason - probably did it from memory and got it wrong!

PR #244, opened 2024-03-02
Let enter key submit forms

bklang:
I noticed that I was not able to type my username & password and then press enter to log in. This snippet of JS is the culprit. I'm sure it was added to fix a problem at some point, but in my testing I can't find any screen where pressing the enter key does something unexpected. I think we should allow default form behavior, and if an issue does come up, figure out a more targeted fix.

### DIFF
--- a/templates/base/go3base.html
+++ b/templates/base/go3base.html
@@ -161,18 +161,6 @@
         })
     </script>
 
-<script type="text/javascript">
-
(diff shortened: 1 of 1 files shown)


aaronoppenheimer:
This always bugged me and I can't for the life of me remember why it worked the way it did.

PR #249, opened 2024-03-02
Update locale files & fix calendar link help

aaronoppenheimer:
I cleaned up the locale text files (local/*/LC_MESSAGES/*.po), and regenerated the .mo files. This also fixed the calendar link help.


aaronoppenheimer:
I'll also add a note in the wiki about generating these in future. Short version: it's using the django standard tools.

bklang (approved):

PR #259, opened 2024-03-04
Calendar feeds & scheduled tasks

aaronoppenheimer:
In looking into calendar feeds, I found bad behavior in the .env file. This PR fixes the default .env so that there are no comments on the same line as settings, which seems to cause the env object to ignore them.

Also added documentation to the README to explain how to work with the scheduled tasks that produce the calendar feeds.

And removed some old infrastructure that was modeled on gig-o 2 but isn't used in gig-o 3.


go3/tasks.py
```diff
@@ -18,20 +18,20 @@
...
-        return HttpResponse()
+# def do_daily_tasks(request):
+#     if int(request.GET.get('key',0)) == settings.ROUTINE_TASK_KEY:
+#         async_task('gig.tasks.delete_old_trashed_gigs')
+#         async_task('gig.tasks.archive_old_gigs')
+#         async_task('gig.tasks.send_snooze_reminders')
+#         async_task('stats.tasks.collect_band_stats')
+#         return HttpResponse('did it!')
+#     else:
+#         return HttpResponse()
```
bklang, line 29: Can this code just be deleted?

aaronoppenheimer: Oops - thought I did. I got rid of the code in urls.py but only commented ot the tasks.py for testing and then didn't delete. d'oh.

bklang:
Thanks! The doc updates are especially helpful. Merging.

bklang (approved):

PR #262, opened 2024-03-05
Stop showing trashed gigs on calendar

tfost:
Attempting to fix https://github.com/Gig-o-Matic/GO3/issues/246. 

Tested this out by  creating two gigs set in the future. Confirmed both appeared on calendar. Moved first gig to trash, then confirmed the first gig no longer appeared on the calendar. 
Images below
Screenshot 2024-03-04 at 11 22 04 PM
Screenshot 2024-03-04 at 11 22 11 PM
Screenshot 2024-03-04 at 11 22 16 PM

### DIFF
--- a/agenda/helpers.py
+++ b/agenda/helpers.py
@@ -77,6 +77,8 @@ def calendar_events(request, pk):
     events = []
     multiband = len(user_assocs) > 1
     for g in the_gigs:
+        if (g.is_in_trash):
+            continue
         gig = {}
         gig['title'] = f'{g.band.name} - {g.title}' if multiband else g.title
         gig['start'] = str(g.date)
--- a/agenda/helpers.py
+++ b/agenda/helpers.py
@@ -78,8 +78,6 @@ def calendar_events(request, pk):
     events = []
     multiband = len(user_assocs) > 1
     for g in the_gigs:
-        if (g.is_in_trash):
-            continue
         gig = {}
         gig['title'] = f'{g.band.name} - {g.title}' if multiband else g.title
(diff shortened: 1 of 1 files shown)


agenda/helpers.py
```diff
@@ -77,6 +77,8 @@ def calendar_events(request, pk):
     events = []
     multiband = len(user_assocs) > 1
     for g in the_gigs:
+        if (g.is_in_trash):
+            continue
```
bklang, line 81: Awesome! Thanks so much for taking this on!

The logic of your fix looks correct, but rather than check each gig as we print them, how about moving this up to the filter line, which is where we pull data out of the database. Perhaps something like `is_in_trash=False` around like 74?

tfost: ah yeah that would be way better, I tunnel visioned on the filtering logic to figure out how all these models worked. Will update with a change to query instead.

tfost: OK - addressed the fix with a new commit, but in doing so I unintentionally created two new commits - one adding the filter to the query and a second removing the filter after the query - still getting used to the GitHub tooling / flows. LMK if its possible to squash these commits on merge, or otherwise I can recreate this as a single commit on a separate branch and re-create the pull request.

bklang: No problem, I can squash when I merge. Thanks!

bklang:
Works great. Thanks so much for the contribution!

bklang (approved):

bklang:
Whoops, forgot to squash. My mistake.

PR #308, opened 2024-03-13
Permissions cleanup

### LINKED ISSUE #304: superuser unable to edit gigs
Superuser should be able to edit gigs for any band but is denied

bklang:
This started as an attempt to address #304 and turned into a permissions refactor.

- Simplify and centralize the checking of administrative permissions - superuser is always a band admin
- Rename template variables to match model attributes for consistency
- Use the UserPassesTestMixin directly instead of cribbing its code

Closes #304

### DIFF
--- a/band/views.py
+++ b/band/views.py
@@ -5,7 +5,7 @@ from django.views.generic.edit import UpdateView as BaseUpdateView
 from django.views.generic.base import TemplateView
 from django.urls import reverse
 from django.contrib.auth.decorators import login_required
-from django.contrib.auth.mixins import LoginRequiredMixin, AccessMixin
+from django.contrib.auth.mixins import LoginRequiredMixin, UserPassesTestMixin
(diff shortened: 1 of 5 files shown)


bklang:
@aaronoppenheimer I believe this is addressed the issue you opened. It tests good for me, but given that it's permissions related, I'd love another set of eyes

aaronoppenheimer:
I'll test it asap. I notice that the superuser is also not able to edit bands from the band page. Basically, any time there's a check for "is this person a member of this band" or "is this person an admin for this band" the answer should be yes for a superuser.

aaronoppenheimer:
I like how you refactored the permissions, and I found that the band views needed a similar overhaul. I moved the permissions checks into the member.helpers file and added what was needed for securing the band, too.

bklang:
Thanks for the assist on this 🙌

PR #313, opened 2024-03-14
Member order

aaronoppenheimer:
Changes so that lists of members show up in alphabetical order.
- display_name was a member property; made it a field (added migration)
- data migration to populate display_name from nickname, username, email address (added migration)
- added ordering to all the calls that return lists of members
- added a signal on member that updates display_name before saving the object
- added a test to show display_name gets updated properly


band/models.py
```diff
@@ -91,15 +92,18 @@ def is_editor(self, member):
...
 
     @property
     def confirmed_assocs(self):
-        return self.assocs.filter(status=AssocStatusChoices.CONFIRMED, member__status=MemberStatusChoices.ACTIVE)
+        return self.assocs.filter(status=AssocStatusChoices.CONFIRMED, 
+                                  member__status=MemberStatusChoices.ACTIVE).order_by(Lower('member__display_name'))
 
     @property
     def confirmed_members(self):
-        return apps.get_model('member', 'Member').objects.filter(assocs__status=AssocStatusChoices.CONFIRMED, assocs__band=self, status=MemberStatusChoices.ACTIVE)
```
bklang, line 102: Did you intend to remove the "active member" filter here? This would mean that someone who is in a band, registers a plan for a gig, and then is marked as inactive, would still show up in the planning for that gig

aaronoppenheimer: Confused - didn't I just move it to the next line?

bklang: Yes - you did. The diff formatting made it hard for me to read, but I see it now. Thanks.

bklang (approved):
Looks good! love the test coverage too - thanks

PR #314, opened 2024-03-14
made the pagination for gigs on the agenda page absurdly long

aaronoppenheimer:
There has been strong pushback against the pagination of gigs on the agenda page. I think it's worth having as an option, but for now it's disabled, practically speaking, by just making the page length very long. I think adding an option is just a matter of checking a preference and either setting the page length to 10 or whatever, or setting it very long like this does.


bklang (approved):
Looks good. While not as compact, I think this will make users happy.

PR #315, opened 2024-03-14
fixed the URL display for the public page

aaronoppenheimer:
The problem was just the url in the band page. Note that the url uses the "URL_BASE" defined in settings, so if you're running it on localhost you'll have to update that (or change it in the url if you're testing).


bklang:
Nice, thanks!

bklang (approved):

PR #318, opened 2024-03-15
Hide cancelled gigs by preference

bklang:
Reverts Gig-o-Matic/GO3#317

Fix the bugs present in the original implementation:

> Tragically, it appears as if selecting "hide cancelled gigs" is also hiding confirmed gigs and, in fact, only showing unconfirmed gigs for me. Whoops!


bklang:
This is the second attempt at solving #202, making it so the preference for hiding canceled gigs works.

The first attempt was buggy, so I reverted it. While fixing the issues related to that, I found that one part of the app referred to gigs as "Cancelled" while another part called them "Canceled". According to [the internet](https://writer.com/blog/canceled-cancelled/), the one-ell version is the preferred spelling for Americans, so I chose that. It has the additional benefit of making any saved preferences still effective, so it's probably the least disruptive change.

@aaronoppenheimer This PR ended up being much larger than expected, so if you have a minute to test it, I'd appreciate it!

bklang:
Ok, this is good to go. It has main branch merged into it, so it includes all the occasional changes that were just merged. This is testing good on my machine. I'll let it sit until this evening in case you want to test it, then I'll merge it.

aaronoppenheimer:
There's an issue when the user has no "gigs with plans" - future_plans() in member.helpers needs to return the "plans"

bklang:
Thanks for catching that. Fixed, and added a test to cover it.

PR #326, opened 2024-03-15
Gigs for occasionals

### LINKED ISSUE #322: Change in user experience for occasional members?
_Expectation:_
As an occasional member I wasn't expecting to see gigs on my home page (usually occasional members were not invited) or to receive emails when gigs are added for that band.

_Current State:_
* I can see approx. every event for a band where I am an occasional member.
* I am prompted to submit my attendance both via the home page and via email.
* I don't think I appear on the event when others in the band look at the gig page?
---
_Example:_
https://new.gig-o-matic.com/gig/262/ - HONK!TX

I show up in the drum section on my view
Screenshot 2024-03-15 at 12 39 00 PM

Also shows up on my home page (along with all other events for that band). These did not show up pre-transition.
Screenshot 2024-03-15 at 1 01 58 PM

Based on screenshots from https://github.com/Gig-o-Matic/GO3/issues/312 I don't believe I show up in the view for others (kinda squinting, but I don't see an "E" in the drum section & number of members doesn't match).

Similarly, I see another occasional member went through and actively "X"ed every event, so I think this may have happened to multiple folks?

---
Anyway, excited to be trying out this new experience! THANK YOU for making this happen. Happy to play around with settings or anything else.

aaronoppenheimer:
Fixed the agenda page so it doesn't show "no occasional" gigs to "occasional" member. Refactored code for generating calendar view and calendar feed to do the same. Added a test for the agenda page.

Should fix #322

### DIFF
--- a/agenda/helpers.py
+++ b/agenda/helpers.py
@@ -69,13 +69,7 @@ def calendar_events(request, pk):
 
     band_colors = {a.band.id: a.colorval for a in user_assocs}
 
-    the_gigs = Gig.objects.filter(
-        (Q(enddate__lte=end) | Q(enddate=None)),
-        date__gte=start,
-        band__in=[a.band for a in user_assocs],
-        hide_from_calendar=False,
-        trashed_date__isnull=True,
-    )
+    the_gigs = [p.gig for p in request.user.calendar_plans]
 
     events = []
     multiband = len(user_assocs) > 1
--- a/agenda/helpers.py
+++ b/agenda/helpers.py
@@ -69,7 +69,14 @@ def calendar_events(request, pk):
 
     band_colors = {a.band.id: a.colorval for a in user_assocs}
 
-    the_gigs = [p.gig for p in request.user.calendar_plans]
+    plans = request.user.calendar_plans
+
+    plans = plans.filter(
+        (Q(gig__enddate__lte=end) | Q(gig__enddate=None)),
+        gig__date__gte=start,
+    )
+
+    the_gigs = [p.gig for p in plans]
 
     events = []
     multiband = len(user_assocs) > 1
--- a/agenda/tests.py
+++ b/agenda/tests.py
@@ -41,6 +41,34 @@ class AgendaTest(GigTestBase):
         # response = c.get(f'/plans/noplans/2')
         # self.assertEqual(response.content.decode('ascii').count("xyzzy"), 9)
 
+    def test_agenda_occasionals(self):
+        joeassoc = self.assoc_user(self.joeuser)
+        janeassoc = self.assoc_user(self.janeuser)
+        janeassoc.is_occasional = True
+        janeassoc.save()
+
+        gig1 = self.create_gig_form(contact=self.joeuser, title=f"xyzzy")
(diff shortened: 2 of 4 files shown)


bklang:
This PR is going to have conflicts with #318. If that one tests OK, I'll merge into this and resolve the conflicts.

aaronoppenheimer:
I ran into mild git-hell when trying to pull while still in the feature branch and ended up rebasing force-pushing, which hopefully didn't change anything except the changelog.

bklang:
Investigating a bit further: it appears that a default plan of "No Plan" is created for occasional members when a new gig is created, even if that member is occasional. Because a plan exists, it shows up on the agenda screens.

aaronoppenheimer:
Occasional users should not see gigs for which occasionals have not been invited - they should not be in their agenda, not in their calendar view or cal feed. They are in the grid, though mostly that's because of how the grid is constructed and I can't decide if it should be changed.

The exception is if the occasional user *has made a plan* for the gig, then the gigs will show up. So, if, for example, a git is "invite everyone" and then an occasional user makes a plan and then the gig changes to "don't invite occasionals" the gig will still show up.

Also, if the occasional member somehow finds the gig (through, say, the grid or by someone sending them the link), they will see on the gig page all of the non-occasional member *and also themselves.*

So, if you're an occasional member and you create a gig to which occasional members are not invited, when you get sent to the gig page after creating it, you will see yourself. If you then go to the agenda or calendar you won't see it. But if immediately after creating the gig you make a plan for yourself on the gig page, that gig will show up everywhere. Maybe that's what you're seeing?

aaronoppenheimer:
> Investigating a bit further: it appears that a default plan of "No Plan" is created for all members when a new gig is created, even if that member is occasional. Because a plan exists, it shows up on the agenda screens.

The code for selecting which gigs appear in the "no plans" section of the agenda is [here](https://github.com/Gig-o-Matic/GO3/blob/55349d298a4207de166098e4b83aa21f648b4f79/member/models.py#L130). If you're occasional, and the gig is "no occasionals", and you have a plan which is NO_PLAN (the default when a gig is created) the gig should be excluded from your list.

bklang:
> > Investigating a bit further: it appears that a default plan of "No Plan" is created for all members when a new gig is created, even if that member is occasional. Because a plan exists, it shows up on the agenda screens.
> 
> The code for selecting which gigs appear in the "no plans" section of the agenda is [here](https://github.com/Gig-o-Matic/GO3/blob/55349d298a4207de166098e4b83aa21f648b4f79/member/models.py#L130). If you're occasional, and the gig is "no occasionals", and you have a plan which is NO_PLAN (the default when a gig is created) the gig should be excluded from your list.

Ok, I'll keep investigating. I'm not sure why I'm seeing these gigs when I shouldn't. Thanks for laying out the rules, that's helpful.

bklang:
Ok, I finally was able to verify the behavior you described. I'm not sure what it was, but it started working correctly after I reset my local environment. I pushed one commit to fix an issue that was broken upstream, unrelated to occasional members.

This looks good - merging, thanks!

bklang (approved):

PR #328, opened 2024-03-16
changed gig contact from pick-list to text field

aaronoppenheimer:
- adds "leader_text" field to gig object
- migrates any existing leaders to leader_text
- changes handling of leader in form & templates

Does not remove the old leader field from the gig object just in case something goes badly in the migration. We can get rid of it later.


bklang (approved):
Tests good for me - Thanks!

PR #333, opened 2024-03-17
Member Order part 2

bklang:
Reverts Gig-o-Matic/GO3#332

The original PR has a side effect that causes members to go missing. It looks like the sorting is the issue. The Django template helper [`regroup`](https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#regroup) requires the input data to already be sorted by the regrouping field. And even though the query appears to try to do that, something is not working.


bklang:
I found the problem. The query is attempting to order by `[section, name]`, which *looks* correct. The plans MUST be grouped by section, or the `regroup` template filter breaks. We ran into this once before in #251.

The "gotcha" here is that ordering by a relation normally orders by its primary key...unless that relation has a defined ordering. Section has a defined ordering, meant to cause sections to always appear in the same order (eg. High Brass -> Low Brass -> Drums). But in at least one case, that ordering is not sufficient to group members into the same sections together because all the sections had the *same order* (`0`) due to the importer creating them without a defined order.

So this is mostly a bug caused by the importer, since it cannot appear in normal use.

Two possible fixes:
* Sort by `section_id` which is always guaranteed to have a unique value (unnecessary if we clean up the data)
* Put a validation on the `Section` model to enforce a unique `order` value (unlikely to happen in the real world, but will catch future mistakes)
* Fix the importer to create sections with a real order

aaronoppenheimer:
Wow, good sleuthing! I guess you’re creating the sections by just parsing the csv file and making a section if you’ve never seen it before? We lose the order, since I’m not actually exporting the section order - when you create them you can just set an order index in the constructor. The band admin can fix the order later.

That said, we should probably have some code someplace that validates the sections - looking at the code in band.helpers where the sections are set up after the order is adjusted, sections are removed, etc - it seems possible that if something went wrong in the front-end and it delivered garbage data to the helper function, things would go wrong. So maybe at the end of the “set_sections” function it’s worth spinning through all of the sections associated with a band to make sure none fell through the cracks.

bklang:
> That said, we should probably have some code someplace that validates the sections - looking at the code in band.helpers where the sections are set up after the order is adjusted, sections are removed, etc - it seems possible that if something went wrong in the front-end and it delivered garbage data to the helper function, things would go wrong. So maybe at the end of the “set_sections” function it’s worth spinning through all of the sections associated with a band to make sure none fell through the cracks.

This is what I'm looking at now. We can put a uniqueness constraint on the `order` field, but then we run into a problem that swapping orders causes a non-uniqueness error. The solution to THAT appears to be a transaction, which is what I'm working on now (not working yet).

bklang:
I think I got it in the latest commit.

aaronoppenheimer:
Great!

I could export the section order numbers in future if that would help.

bklang:
Alright, I think this is ready to go. It now has database protection around the section ordering, and there is code to auto-assign values if one is not provided.

I appreciate the offer to export section ordering. If you want to start exporting the band details as JSON, then I'll take you up on it. If we go that route, we can also preserve the other band profile info. It's up to you - if you want to put that work in on the exporter, I'll upgrade the importer.

PR #334, opened 2024-03-18
Remove dead code

bklang:
- [ ] Needs test coverage


aaronoppenheimer:
I think this is actually overcome by events - my refactoring of the calendar-gigs-calculation moved the decision-making into the member model (member.calendar_plans) and seems to do the right thing.

bklang:
You're right, and the tests back it up. Switching this to just remove some dead code

Issue #347, opened 2024-03-20
when changing email address, confirmation should go to new address not old


aaronoppenheimer:
Changing email sends a confirmation to the new address - except go3 is sending it to the old address!

aaronoppenheimer:
fixed in #348

bklang:
Fixed in #348

PR #348, opened 2024-03-21
Lowercase emails

aaronoppenheimer:
Added various bits so emails are properly lowercased:
- when creating a member object (using either the membermanager or when saving a member object)
- when inviting members
- added login authentication bit to compare to database without case sensitivity
- added a test case or two

Also
- fixed bug in email-address change process so confirmation goes to new address, not old

Plus
- added some helpful debugging things: ability to disable captcha from settings, printing out email body when email is disabled

### DIFF
--- a/member/models.py
+++ b/member/models.py
@@ -177,6 +177,12 @@ class Member(AbstractUser):
         return EmailRecipient(name=self.username, email=self.email,
                               language=self.preferences.language) # pylint: disable=no-member
 
+    def save(self, *args, **kwargs):
+        """ when creating members with a form, the usermanager isn't used so we have to override save """
+        self.email = self.email.lower()
+        # then super
+        super().save(*args, **kwargs)
+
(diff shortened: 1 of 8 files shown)


member/models.py
```diff
@@ -247,4 +253,4 @@ class EmailConfirmation(models.Model):
     language = models.CharField(choices=LANGUAGES, max_length=200, default='en-US')
 
     def as_email_recipient(self):
-        return EmailRecipient(email=self.member.email, language=self.language)
+        return EmailRecipient(email=self.new_email, language=self.language)
```
bklang, line 256: Does this also fix #347 ?

aaronoppenheimer: It does!

aaronoppenheimer:
The "add a backend to the validation system" was a new idea for me - pretty cool.

bklang (approved):
LGTM! Tested good locally

PR #352, opened 2024-03-21
Switch to file backend for email in development

bklang:
The previous implementation of the `EMAIL_ENABLE` flag was good, but it only caught instances where the email was triggered by that particular code path. Notably, it left out instances where email was sent by the password reset mechanism. This switches the email backend out for a file-based backend when Django is running with `EMAIL_ENABLED=False`.  Look for sent emails to appear in the `tmp/` folder inside of the application root directory.


bklang:
@aaronoppenheimer tagging you for visibility

Issue #354, opened 2024-03-21
Band description field too short


bklang:
The old site allowed much longer descriptions. Imported bands are getting truncated

aaronoppenheimer:
I think there was actually no limit before. What's a useful limit?

bklang:
Fixed!

PR #357, opened 2024-03-21
removed the limit on number of characters in some of the fields.

aaronoppenheimer:
Addresses #354 

Interestingly, putting a limit on a textfield field should not affect the data as stored in the database, but only sets a limit on the textfield widget used in forms. So when importing a band, a long description seems like it should be allowed - but when you try to look at it in the admin page or band edit page, it gets truncated. I guess? In any case, taking the limits out seems to fix the problem.

@bklang There are migrations here - please test this by setting a band to having a long description, member links, etc, and then migrate to make sure nothing gets lost. As far as I can tell, the migration doesn't harm any existing content.

(ETA: I tested it, natch, but I don't trust myself)


bklang (approved):
Looks great to me! Thanks!

PR #359, opened 2024-03-21
made it possible to invite alumni to a band

aaronoppenheimer:
If you invite an "alumni" member, their status switches back to "invited" and they can accept the invitation.


aaronoppenheimer:
This works but there are some corner cases. If a member is CONFIRMED, then the leave and are ALUMNI, and then are INVITED but they decline the invitation, the invitation is deleted but the member is in limbo - the assoc is still in the "invited" state but since there's no invitation, the profile page doesn't know what to show so it shows nothing.

If there's an existing assoc for a band, which means at some point the member was a member, and it's declined, I guess make it go back to "alumni"? Maybe there needs to be a "this member was once an alumni" to be sure?

aaronoppenheimer:
Similarly, if someone leaves and then requests to rejoin, they become PENDING. If the admin rejects them, the assoc is deleted and they are no longer ALUMNI. Boo.

aaronoppenheimer:
Right now the status of an assoc is:

- UNCONFIRMED (which is used only as the default when an assoc is created - probably don't need it at all)
- CONFIRMED (which means they are a regular member)
- INVITED (which means they have been sent an invitation)
- PENDING (which means they have requested to join)
- ALUMNI (which means they used to be a member but left or quit)

So the problem is if you're ALUMNI and switch to INVITED or PENDING and reject the invite or are not accepted into the band, what should happen? You should go back to ALUMNI, I think.

So, I will adjust things this way:
- When an assoc is saved with the status set to CONFIRMED it will have an "alumni" flag set.
- When a member leaves a band, status will be set to UNCONFIRMED but alumni flag will still be set.
- If a member is INVITED or PENDING and they don't end up joining, if the alumni flag is set they'll just go back to UNCONFIRMED
- If a member is INVITED or PENDING and they don't end up joining, if the alumni flag is *not* set, the assoc will be deleted.

I think that will be rational behavior. This is all new to Go3 because there was no notion of "alumni" before.

aaronoppenheimer:
HAS MIGRATIONS

aaronoppenheimer:
OK - merged in migrations from main, remade migrations for this branch. Tested data migration and it seems ok!

bklang:
Sorry for the delay on this, I wanted to test it and think through it.  I modeled all the statuses I could find and test in the code, here is what I have found:

Band Assoc Flowchart

A couple of observations:
* The `is_alum` boolean is never false
* The `AssocStatusChoices.INVITED` status is unused anywhere
* It is not possible to unilaterally request to join a band for non-alums (was this ever possible?)

bklang:
The above concerns are not really a problem, and merging this PR helps people using the site. I'll leave it as notes to our future selves and go ahead and merge this. Thanks for addressing it!

bklang:
I found the way for a Gig-O user to request joining a band. I'll update the flowchart for that

aaronoppenheimer:
The INVITED status is used when a member is an alum but it then invited back to the band - [here](https://github.com/Gig-o-Matic/GO3/blob/75258500c61f041e3f6222eeb523962ea1f5de58/member/views.py#L263). It's not really used as a filter - for example, when the band is displaying its invited members it just looks for "invite" objects. But it seems important (or at least complete) to put an assoc into the "invited" state.

Go2 created an assoc when a member was invited, which caused headaches, so Go3 doesn't bother until an invitation is accepted.

Issue #242, opened 2024-03-02
Incorrect end dates on event series


bklang:
When using "Add A Series Of Copies" checkbox, the backend creates several copies of the new gig by adjusting the date forward. If the user does not provide an end date, then all copies in the series will have the same end date, which is defaulted to the start date of the first event in the series. 

Reproduction:
* Create a new event with a start date of March 2 and no end date
* Check box to add a series and specify 4 weekly events
* Save the event
* 4 new events are created with start dates of March 2nd, 9th, 16th, and 23rd, but all 4 events have an end date of March 2nd.

Also check the end times, they may not be correct (could be related to #264?)

aaronoppenheimer:
This was fixed at some point - if you don't provide an end date for a gig, there's no end date at all and same for copies.

PR #379, opened 2024-03-24
Time cleanup

aaronoppenheimer:
I think this is ready for review. It's not done - I still need to make sure times are shown properly on the agenda page, and would like to revamp the form validation. But this should at least stop screwing up blank gig times.


aaronoppenheimer:
I think I need a data migration for existing gigs - if the date is not set to midnight that means there was a true start time merged in, so need to set the "has_call_time" flag, and same with the other two. And can probably discern all-day gigs from the old times too.

aaronoppenheimer:
OMFG. I learned why when you save a gig without a time, you end up with a time of 12:56:

`
from datetime import datetime
import pytz
d = datetime(2024,4,1,12,0,tzinfo=pytz.timezone('US/Eastern'))
print(d.astimezone(pytz.utc).astimezone(pytz.timezone('US/Eastern')))
`
So we took a time in eastern, converted to UTC and back, and we get:
`2024-04-01 12:56:00-04:00`

Blah. So in any case, gig migration 0020 is a data migration that looks at gigs date, setdate, and enddate, and (hopefully) properly sets the new flags "is_full_day" "has_call_time" "has_set_time" and "has_end time". I have tested it by putting in a bunch of gigs with various kinds of times and running the migration and checking, but I'm not yet convinced I have checked every case, so will keep testing.

bklang:
Wow, what a surprising result! This jogs my memory about a [post I read a couple of days ago](https://stackoverflow.com/questions/60996205/python-datetime-timezone-conversion-off-by-4-minutes)

I don't really understand the reason for the behavior, but apparently you are supposed to use the `localize()` method: 

```Python
d = datetime(2024,4,1,12,0)
tz = pytz.timezone('US/Eastern')
tz.localize(d) # '2024-04-01 12:00:00-04:00'
```

aaronoppenheimer:
Huh - so maybe it somehow loses a minute per hour of timezone difference. I'm going with rounding error. Anyway, I'll see if the migration works. Once that works I think it'll be easy to add per-gig timezones. For some definition of "easy."

aaronoppenheimer:
@bklang OK, I think this is working properly. I tested with bands based in US/Eastern, US/Pacific, and UTC. The migration mostly just sets the "all day" and "has call time" etc. flag, but for full-day gigs is also strips out the time and just sets it to midnight (local for the band). If we're nervous, I believe we can *not* change the times at all and still get the right behavior since the times aren't actually shown for full-day gigs. That would make it easier to back out if needed. Or the migration could squirrel away the old datetimes just in case, so the migration would be truly reversible. What do you think? We won't need those old migrations in future but might be nice to have..?

gig/views.py
```diff
@@ -57,15 +57,17 @@ def get_context_data(self, **kwargs):
         context['user_has_manage_gig_permission'] = has_manage_gig_permission(
             self.request.user, self.object.band)
 
-        tz = pytz.timezone(self.object.band.timezone)
-        start_time = self.object.date.astimezone(tz).strftime("%H:%M")
-        context['gig_has_start_time'] = (start_time != "00:00")
+        # tz = pytz.timezone(self.object.band.timezone)
+        # start_time = self.object.date.astimezone(tz).strftime("%H:%M")
+        # context['gig_has_start_time'] = (start_time != "00:00")
```
bklang, line 62:
```suggestion
```

gig/views.py
```diff
@@ -156,8 +158,8 @@ def get_context_data(self, **kwargs):
     def get_success_url(self):
         return reverse('gig-detail', kwargs={'pk': self.object.id})
 
-    def clean_date(self):
-        pass
+    # def clean_date(self):
+    #     pass
```
bklang, line 162:
```suggestion
```

bklang (commented):
Looks good. I've added a few suggestions to prune some dead code. Continuing to test this locally

bklang:
Most everything is testing good so far, but I did notice this discrepancy in the iCalendar feed:

Screenshot 2024-03-28 at 10 37 34 PM

The gig is scheduled in Gig-O without any time specified, it should be reflected in iCalendar as an all-day gig.

Screenshot 2024-03-28 at 10 38 23 PM

bklang:
The last commit I pushed should address the iCalendar issue, at least for all-day/multi-day gigs.

bklang:
My testing looks good too. There are still a few issues not directly caused by this PR, but adjacent. I'll list them here and propose we fix them in follow-up PRs, since this one is already getting long in the tooth:

- [ ] Data issues: because of the [bug related to recurring event creation](#242), we have 58 events (and counting) in the DB that have end dates *before* start dates. This causes parse errors in the icalendar feed. At least on Mac, this makes those particular events disappear from the calendar.
- [ ] Events which have no start time are appearing at midnight, both in the Gig-O calendar and in the icalendar feed

> The migration mostly just sets the "all day" and "has call time" etc. flag, but for full-day gigs is also strips out the time and just sets it to midnight (local for the band). If we're nervous, I believe we can _not_ change the times at all and still get the right behavior since the times aren't actually shown for full-day gigs. That would make it easier to back out if needed. Or the migration could squirrel away the old datetimes just in case, so the migration would be truly reversible. What do you think? We won't need those old migrations in future but might be nice to have..?

I like the idea of not wiping out the times. That seems minimally destructive. I'll update the migration for that and re-test.

aaronoppenheimer:
Events that have no start time are considered full-day events and so should show up at midnight (band-local).

For the events that have broken times, is a migration the best way to fix it? I guess it wouldn't be hard to just fix it in the django shell. I also think about putting in signals that just check for valid times and rationalize them if they're broken, but that seems like the wrong way to go.

Another thing to fix - duplicate events should have their times copied over into the new event. I'm looking into that one now.

bklang:
> Events that have no start time are considered full-day events and so should show up at midnight (band-local).

Why not treat these as full-day gigs (eg. "time unknown, but date known"), rather than midnight? That's the behavior you would see in most calendaring apps.

> For the events that have broken times, is a migration the best way to fix it? I guess it wouldn't be hard to just fix it in the django shell. I also think about putting in signals that just check for valid times and rationalize them if they're broken, but that seems like the wrong way to go.

Yep, this needs to be a migration - a one-time fix that hopefully won't be necessary after these changes go in.

bklang:
> * Some bands are creating gigs with a Set time, but no Call time. Due to the history of Gig-O, Set time is treated as the most important time. When saving a gig without a Call time, but a Set time provided, we should copy the Set time to the Call time.

Thoughts on this issue I just found? ^

bklang:
> > Events that have no start time are considered full-day events and so should show up at midnight (band-local).
> 
> Why not treat these as full-day gigs (eg. "time unknown, but date known"), rather than midnight? That's the behavior you would see in most calendaring apps.

I just realized we are mostly saying the same thing. I fixed this so gigs without a time now show up like this:
Screenshot 2024-03-29 at 6 05 27 PM

Treated as a full-day gig, but no longer showing up at midnight. It's just a time-less entry for that date.

aaronoppenheimer:
> Treated as a full-day gig, but no longer showing up at midnight. It's just a time-less entry for that date.

Oh, I see what you meant - great.

aaronoppenheimer:
> > * Some bands are creating gigs with a Set time, but no Call time. Due to the history of Gig-O, Set time is treated as the most important time. When saving a gig without a Call time, but a Set time provided, we should copy the Set time to the Call time.
> 
> Thoughts on this issue I just found? ^

Is this referring to the calfeed? I think that's correct - if a gig has a set time but not a call time, it should show up on the calendar at the set time. If it has an earlier call time, it should show up at the call time. Is that what you meant?

bklang:
> Is this referring to the calfeed? I think that's correct - if a gig has a set time but not a call time, it should show up on the calendar at the set time. If it has an earlier call time, it should show up at the call time. Is that what you meant?

Yep - that is what I meant. And it's not currently happening. Fix coming up.

bklang:
@aaronoppenheimer Ok, I'm happy with this and I think it's ready to merge. Do you want to give it one more look before I push the button?

aaronoppenheimer:
Looking!

Full-day events show up as colored bars rather than boxes with colored dots. If the color is "no color" the box is white-on-white-with-white-text. It's there but invisible. Maybe should be light gray or something?

gig/forms.py
```diff
@@ -130,6 +130,14 @@ def _mergetime(hour, minute='', zone=None):
             self.cleaned_data['has_set_time'] = not set_time is None
             self.cleaned_data['has_end_time'] = not end_time is None
 
+            if not call_time:
+                if set_time:
+                    # if set time is set, but call time is not, default call time to match set time
+                    call_time = set_time
```
aaronoppenheimer, line 132: I think this is OK but I find it uncomfortable. I know when we're playing but I don't care when you show up - why force a call time in there? It might be the right thing to do, just unexpected (for me).

aaronoppenheimer:
Other than the color thing, seems good to go!

gig/forms.py
bklang: The reason I went this way is because many parts of Gig-O expects `date` to be *the* start datetime for the event, and ignore the settime. The most visible places were in the agenda calendar view and the icalendar feeds, but I think there are other places as well. Maybe we could fix this with accessor methods on the model? But I was trying to minimize the amount of extra complication on this already large PR. 

I also would prefer to leave the input as the user intended it. Maybe in a future PR we can fix all the places that depend on `date` and then put this back to being truly optional?

bklang:
> Full-day events show up as colored bars rather than boxes with colored dots. If the color is "no color" the box is white-on-white-with-white-text. It's there but invisible. Maybe should be light gray or something?

Nice catch, I missed this. Will update the color.

aaronoppenheimer:
That one was a journey! Thanks Ben!

PR #380, opened 2024-03-24
changed content type for calfeed to text/calendar

aaronoppenheimer:
@bklang The ical validator think the content was OK but when pulling from URL it complained about the content-type. This will make it happy - any reason not to?


bklang (approved):
Yeah, good catch!

PR #391, opened 2024-03-30
Long comment

aaronoppenheimer:
Currently an extra-long gig plan comment causes problems since the gig model has a limit of 200 characters. Rather than remove that limit, which is think is a good thing given the point of the comment field, here I've limited the number of characters the field allows and put in a truncation on the back end.

The only issue now is that the placeholder text "200 characters" can't be translated normally - because the javascript that produces it is loaded as a script, it doesn't go through the translation processor. Maybe there's a way to do that, or maybe it's better to just {% include %} the js file within the larger <script> content on the html page, which would then (I think) put it through the translator. Any issue with doing it that way?


bklang:
I'm looking at the translation question. The Django docs seem to have some advice on this: https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#internationalization-in-javascript-code

aaronoppenheimer:
That's cool. I can give it a try unless you're already deep into it.

bklang:
> That's cool. I can give it a try unless you're already deep into it.

Nope, not yet. I'm working on some other things this morning

bklang:
> There's a note on performance in the documentation [here](https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#note-on-performance) - what do you think we should do?

At the scale we are at, I don't think this is going to be a problem for us. We'd get a bigger lift from moving static assets to being served from outside the app. We can revisit this if and when we decide to do that.

bklang (approved):

PR #398, opened 2024-03-31
fixes for better handling of times on the edit form

aaronoppenheimer:
Decouples the gig object from the rendering of the form for clearer handling of end dates and better toggling between all-day and not-all-day events.

### DIFF
--- a/gig/views.py
+++ b/gig/views.py
@@ -187,9 +187,6 @@ class UpdateView(LoginRequiredMixin, UserPassesTestMixin, generic.UpdateView):
         # we need to do this because there are some form fields that do not exist in the object,
         # so we need all of the initial values from the object so we can get at them
         # from the template.
-        # XXX
-        # kwargs['initial'] = forms.models.model_to_dict(self.object)
- 
         return kwargs
 
     def get_success_url(self):
--- a/gig/views.py
+++ b/gig/views.py
@@ -177,7 +177,7 @@ class UpdateView(LoginRequiredMixin, UserPassesTestMixin, generic.UpdateView):
         context = super().get_context_data(**kwargs)
         context['timezone'] = self.object.band.timezone
 
-        SetupFormTimes(context,self.object)
+        setup_form_times(context,self.object)
 
         return context
 
--- a/gig/views.py
+++ b/gig/views.py
@@ -146,7 +146,7 @@ class CreateView(LoginRequiredMixin, UserPassesTestMixin, generic.CreateView):
         return result
 
 
-def SetupFormTimes(context,object):
+def setup_form_times(context,object):
     context['date'] = object.date
     if object.is_full_day:
         if object.enddate and object.enddate.date() != object.date.date():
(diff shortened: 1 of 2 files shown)


gig/views.py
```diff
@@ -146,6 +146,24 @@ def form_valid(self, form):
         return result
 
 
+def SetupFormTimes(context,object):
```
bklang, line 149: Small nit, but method names typically are underscored, while classes are initial-caps
```suggestion
def setup_form_times(context,object):
```

gig/views.py
```diff
@@ -158,6 +176,9 @@ def test_func(self):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context['timezone'] = self.object.band.timezone
+
+        SetupFormTimes(context,self.object)
```
bklang, line 180:
```suggestion
        setup_form_times(context,self.object)
```

gig/views.py
```diff
@@ -166,14 +187,9 @@ def get_form_kwargs(self, *args, **kwargs):
...
-        
-        # if this is not a full day event, or the date and enddate are the same,
-        # make the enddate widget on the form blank
-        if not self.object.is_full_day or \
-            self.object.enddate and self.object.date.date() == self.object.enddate.date():
-            kwargs['initial']['enddate'] = None
-        
+        # XXX
+        # kwargs['initial'] = forms.models.model_to_dict(self.object)
+
```
bklang, line 192:
```suggestion
```

gig/views.py
```diff
@@ -195,28 +211,40 @@ def form_valid(self, form):
...
         if not self.request.user.is_authenticated:
             return self.handle_no_permission()
         gig = get_object_or_404(Gig, id=self.kwargs['pk'])
         return gig.band.is_editor(self.request.user)
 
+    def get_context_data(self, **kwargs):
+        self.original_gig = Gig.objects.get(id=self.kwargs['pk'])
+        context = super().get_context_data(**kwargs)
+
+        SetupFormTimes(context, self.original_gig)
```
bklang, line 230:
```suggestion
        setup_form_times(context, self.original_gig)
```

gig/views.py
```diff
@@ -195,28 +211,40 @@ def form_valid(self, form):
...
-        kwargs['initial'] = forms.models.model_to_dict(gig_orig)
+        kwargs['initial'] = forms.models.model_to_dict(self.original_gig)
         # ...but replace the title with a 'copy of'
         kwargs['initial']['title'] = f'Copy of {kwargs["initial"]["title"]}'
         return kwargs
 
     def get_band_from_kwargs(self, **kwargs):
-        # didn't have the band from the request args, so pull it from the gig
-        return get_object_or_404(Gig, id=self.kwargs['pk']).band
+        return self.original_gig.band
```
bklang, line 247: It seems like this method's behavior no longer matches its name. Do we even still need it?

aaronoppenheimer: We do need it but the name could be better. Will fix.

bklang (approved):
Looks good, tests good. Thanks for tackling this!

PR #399, opened 2024-03-31
fixed flicker and updated fontawesome use

aaronoppenheimer:
updated how we use fontawesome to bring in their django library which does a better job of caching things


bklang (approved):

bklang:
Nice, this should make page loads more reliable too. Thanks!

Issue #321, opened 2024-03-15
Admins should be able to edit archived attendance


nocturne-mit:
It appears band admins still cannot edit the attendance for a gig once it's archived, the day after the gig end date?

If there's no architectural reason making this particularly difficult, it would be very handy if a band admin could still edit attendance for at least a few days after the gig end date. (Being able to edit indefinitely would also be fine by me.)

If editing archived gigs is too difficult, would it possibly make sense to defer archival for a small number of days, to allow this kind of post-facto edit before archiving happens?

Our band relies on the GigO attendance info for payout purposes, so it would be helpful if we can make last-minute (or after-the-fact) edits to gig attendance, to record last-minute changes (or to note whether a "probably" actually showed up).

On a related note, I love the change from GO2 where archived gig attendance shows up in GO3 with color icons instead of monospace words - it's MUCH easier to skim the attendance of 20 band members this way.

nocturne-mit:
I don't have a strong opinion about automatic gig freezing, I think I'm fine either way.

bklang:
Thanks for the suggestions @nocturne-mit . Aaron addressed this in #401

PR #401, opened 2024-04-01
Improved archving

aaronoppenheimer:
Updated archive system so that a gig isn't "archived" - which is to say, frozen so the gig plans can't be edited - for five days. Until then, gigs that have passed are in the archive page but the plans can still be edited. This is to address #321 for cases where, say, a leader can go back to a gig and update who was actually there before it gets frozen.


bklang (approved):
LGTM. Tested locally by removing the archive flag from a gig. It still appeared in the band archive, and was editable.

PR #403, opened 2024-04-01
Show gig time agenda

aaronoppenheimer:
There was a preference to show gig times on the agenda page, but it was unused. I added a UI for it on the preference-edit page (which really needs an overhaul); added the code to actually show the time according to the preference; updated the language files; and noticed the the preference edit page wasn't translating labels so I fixed that while I was in there.


aaronoppenheimer:
Something is weird with (no surprise) timezones - if I make gigs and then change the timezone, the time shown on the agenda page might or might not be shown in local time, but when I refresh it's in UTC. It only happens sometimes, which is especially weird - it's correct once and then reverts to UTC when I refresh the page...?

aaronoppenheimer:
OK, there is definitely a weirdness when you change the timezone for a band and then look at, say, the agenda page. This is because the page doesn't know what timezone to display times in. As long as you do everything within one timezone, it will work fine, but if you change a band's timezone, it becomes unclear what timezone the times are displayed in.

I think this should be merged as-is, which will work fine for most cases. And we should bite off a timezone overhaul, which would give each gig a timezone, and give the user a timezone (which we could set automatically, maybe using something like [this](https://github.com/adamcharnock/django-tz-detect)).

templates/member/member_detail.html
```diff
@@ -185,6 +185,12 @@
                     </div>
                 </div>
                 <div class="card-body">
+                    <div class="row">
+                        <div class="d-none d-sm-inline col-sm-3">{% trans "Show gig time on schedule" %}:</div>
+                        <div class="d-none d-sm-inline col-sm-9">{% include "base/tf.html" with value=member.preferences.agenda_show_time %}</div>
+                        <div class="col-12 d-sm-none">{% trans "Show gig time on schedule" %}: {% include "base/tf.html" with value=member.preferences.agenda_show_time %}</div>
+                    </div>
+                    <br>
```
bklang, line 193: Interesting that it seems like the code was there, but the preference was not exposed. 

Do you know the use case for *not* showing gig times? Would it make more sense to just always show them?

aaronoppenheimer:
Yeah, clearly it was half built. I don't know if there's a reason to turn it off. I am thinking about overhauling the agenda page options to make them clearer and allow splitting out canceled gigs and a few other things.

For now, maybe we should just turn it on? Get rid of the UI and just assume everyone wants it? Run a migration to turn the option True for everyone? Just notify them in the MOTD that the option exists now?

bklang:
> For now, maybe we should just turn it on? Get rid of the UI and just assume everyone wants it? 

I like this approach. Too many options makes for a worse user experience, and more complexity for us to test and maintain.

member/models.py
aaronoppenheimer, line 1: The reason to add labels in the model file is that it lets Django generate "edit" views automatically - so the translations have to go here. Otherwise we have to make a view to do this just to set the labels.

PR #406, opened 2024-04-02
extended maximum session time to 1 year

aaronoppenheimer:
Django default is 2 weeks. Old gig-o never expired. Any reason not to extend to a year?


bklang (approved):

bklang:
Agree. These are low stakes cookies, we should be user friendly

PR #407, opened 2024-04-03
Recaptcha forms

aaronoppenheimer:
Fixed or added captcha_enable functionality to signup, password reset, and new-band-request forms.


aaronoppenheimer:
So...this is failing because the captcha is failing. It seems like we should disable the captcha but I don't know what environment file is used, if any, during testing. Is there a way to specify one for this purpose, and we can disable it properly? Otherwise I can hack in some workaround (or remove the test, it's not that critical).

bklang:
Probably the easiest thing to do is to provide whatever environment variables you need to disable the ReCAPTCHA to Github: https://github.com/Gig-o-Matic/GO3/blob/main/.github/workflows/testing.yml#L5
Maybe `CAPTCHA_ENABLE: False` is enough?

aaronoppenheimer:
OK, I figured out what we can do. We can set up an alternative .env file for use in testing that just turns the CAPTCHA_ENABLE false, and then in the testing.yml file, just direct the system to use the alternate.

aaronoppenheimer:
Oh, I guess this will fail tests until it is merged in. That's a bummer - should have split it into two PRs.

aaronoppenheimer:
Oh, I didn't know you could do that. Easier than my option. Will back it out.

bklang (approved):

PR #412, opened 2024-04-04
adjusted schedule page so canceled gigs always show up in the "has plans" panel

aaronoppenheimer:
Adjusted schedule page so canceled gigs always show up in the "has plans" panel. They can be hidden completely with the preference item, but even if there are no plans for a gig that gets canceled it shouldn't show up in the "weigh in" panel because there's no way to do that and WHO CARES? it's canceled! But it shouldn't just disappear.


bklang (approved):
Nice. Thanks!

PR #413, opened 2024-04-05
working on updating the gig email

aaronoppenheimer:


aaronoppenheimer:
Go3 has a system built in where editing a gig lets you look at what changed and put a note about it in the notification email. It was easier to do before we had 'is full day' and other date/time features, so I've pulled the details out in favor of just "the date/time changed" and we can readdress it later.

aaronoppenheimer:
@bklang I'd like to make sure you're OK with this plan before I go through and fix all the tests. The way it's set now, if anything changes about the gig date/time, the "changelog" in the email just says "Date/Time (see below)" instead of being explicit about the change. Seem ok, at least for now?

bklang:
Sorry, I just noticed this was waiting on me. 

Your suggestion sounds reasonable to me. The most important thing is that we call out what has changed, and that the new information can be seen.

PR #439, opened 2024-04-21
We are out of beta for GO3! 🎉

bklang:


bklang:
@aaronoppenheimer FYI: you can use Markdown in MOTD messages now

Issue #445, opened 2024-04-22
fully qualified band "Website" URL looks like it should work but does not


nocturne-mit:
When setting up my newest band, I put a fully qualified URL (e.g. "https://foo.org/") in the band's "website" field.
This HTTPS URL is displayed verbatim as text when you view the band-info page, but if you click on the link, it's actually a hyperlink to "http://https//foo.org/".

Feel free to solve this however seems best, but in case suggestions are helpful, you could take one of these approaches:
1. Forbid entry of a URL that specifies the URL scheme (i.e. "https:"), but this means people can't specify https:// URLs for sites that don't have working HTTP-to-HTTPS redirection
2. On the band-info display page, only prepend "http:" when the URL doesn't already start with a URL scheme (flexible for bands, but maybe harder to code)

aaronoppenheimer:
should be addressed in #446

PR #446, opened 2024-04-22
better handling of band website url

aaronoppenheimer:
Instead of adding "http://" to the url in the template, I clean it up in the form when the band is edited.


bklang (approved):

bklang:
Closes #445

PR #450, opened 2024-04-23
Various task queue settings for safer operation

### LINKED ISSUE #449: Email storm?
SFMA sent thousands of emails related to gig changes. The band admin says she only hit save a few times, and that way more emails were sent than the number of times she saved changes. Needs investigation.

bklang:
* Allow longer-running tasks to get through emails to large bands
* Fix typo in configuration to not repeat failed jobs (`ack_failures`)
* Ensure we do not retry jobs which timeout

Fixes #449

### DIFF
--- a/go3/settings.py
+++ b/go3/settings.py
@@ -260,8 +260,9 @@ if ROLLBAR_ACCESS_TOKEN:
 Q_CLUSTER = {
     "name": "DjangORM",
     "workers": 1,
-    "timeout": 120, # Allow for longer task runs, particularly emails to large bands
-    "retry": 180, # Always ensure this is larger than timeout, or tasks will duplicate!
+    # Set timeout ridiculously high until we have a solution: https://github.com/Gig-o-Matic/GO3/pull/450#issuecomment-2072130002
+    "timeout": 600, # Allow for longer task runs, particularly emails to large bands
+    "retry": 660, # Always ensure this is larger than timeout, or tasks will duplicate!
     "max_attempts": 1, # Prevent duplicate task runs in the case of a timeout or error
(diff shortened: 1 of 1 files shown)


aaronoppenheimer:
Don't forget [this one](https://github.com/Gig-o-Matic/GO3/blob/0334b2a310e2f6b579430c412b16fe3aa6c324ef/lib/email.py#L17) though maybe we don't need it if the global setting is fixed.

bklang:
I assume you're referring to `ack_failure` vs. `ack_failures` - if so, thanks for the callout. Confusingly, the global config is plural while the individual method call is singular. 

This PR isn't totally ready yet though. Tasks are still being retried after a timeout. I found another undocumented config option that I am testing now.

aaronoppenheimer:
> I assume you're referring to `ack_failure` vs. `ack_failures` - if so, thanks for the callout. Confusingly, the global config is plural while the individual method call is singular.

D'oh. Now we know how the typos get in there.

bklang:
Looks like this is [still a known issue](https://github.com/django-q2/django-q2/issues/173)

I'm not sure what to do right now. Some options:
* Reduce batch size and/or increase timeout to try to avoid timeouts in the first place (imperfect)
* Fix the issue upstream in django-q
* Add our own timer and raise before the task times out, to force a failure (may not be reliable)

bklang:
Since I've got email disabled globally right now, I'm going to opt for the ridiculously high timeout. Worst case, the queue will get blocked and email won't go out. But at least we won't be spamming peoples inboxes.

After that, I'm going to keep working on a solution for the upstream django_q.

PR #451, opened 2024-04-23
when emailing about a gig, set up async tasks in chunks

aaronoppenheimer:
Not very complicated. I don't know if 20 is the right number but I don't think there's much of a penalty.


aaronoppenheimer:
Don't think we need this - it was to solve a problem that seems to have gone away.

PR #472, opened 2024-05-11
added django debug toolbar

aaronoppenheimer:
Not sure if we want to add this, but it seems pretty useful - check it out.


aaronoppenheimer:
interesting but not needed in codebase

PR #477, opened 2024-05-21
Stats fixes

aaronoppenheimer:
Adds a system-wide "emails sent" metric and cleans up the existing metrics so they actually work.


aaronoppenheimer:
The current stats collection is broken - we should just delete all of the metrics/statistics that exist and start from scratch with these once it's accepted.

PR #488, opened 2024-06-05
Split up function that creates and returns member plans, so we're not…

aaronoppenheimer:
… going through the creation process all the time. It's not a giant database hit but I think we want to control when/where stuff is inserted into the database.

Basically, any time we were requesting a list of plans from a gig, we were checking to see if there are any members that don't have plans and creating them. However, we don't always need to do that - we only really need to do it when creating a gig or when adding a member. So I split the function so "Gig.member_plans" just returns the plans, and "Gig.update_plans()" does the work to make new ones and is called from the Gig and Assoc "save" signals.

I doubt this fixed the problem I'm chasing, which is that bands seem to not be sending the right number of emails all the time, but there's no reason to recalculate which plans are missing every time we want a list of plans, especially since most of the time there won't be any missing.

I also think we should control when the plan objects are created - we don't necessarily want it to be done in an asynchronous task, which was possible in the old code.


aaronoppenheimer:
Hm. I'm not really sure this is necessary - maybe better to leave as-is, since it's working fine.

aaronoppenheimer:
I don't think we need this.

PR #494, opened 2024-06-16
Schedule formatting

aaronoppenheimer:
Adds feature to let you format the agenda page in different ways: by needs-answer, by band, or single list


aaronoppenheimer:
HAS MIGRATION

aaronoppenheimer:
@bklang if you get some time, can you pull this one down and try it? It's a big enough change that I'm nervous releasing it into the wild! (the new feature only works if you set the 'is_beta_tester' flag on your member object for now)

bklang:
Very cool! I'm playing with this to get a feel for how it works.
I am running into a bug where the drop-down for selecting the view on the preferences page is intermittently appearing & disappearing. Looking into that.

aaronoppenheimer:
I think that's to do with the 'is beta tester' flag - I saw that behavior when was testing the flag, setting/unsetting while looking at the preferences page. If I forced a reload it would do the right thing, or if I left and came back. But just reloading the page after changing the flag didn't do it.

It's likely there's something in there I don't understand about dynamically changing the fields in the class-based views!

bklang:
I've just pushed up a commit that I believe resolves the issue with the mysteriously disappearing preference selector.

aaronoppenheimer:
Was the problem before that the method I was using gets called when the view is instantiated, and stays around so the beta flag isn't reevaluated with every serving of the view?

aaronoppenheimer:
I'm still seeing the same (or similar) problem:

- Start server
- In admin pages, set user as betatester
- Open user preferences -> see the agenda layout picker
- Cancel out of preferences
- In admin pages, set user as not a betatester
- Reload the user preferences page -> still see the agenda layout picker

Setting breakpoints in the view, it looks like when we're adding the 'agenda layout' field, when we reinstantiate the view it's still there. In my version I explicitly removed it if it was there but user was not a betatester, though I'm not sure why I had to do that. The view is persistent somehow..? I put in an __init__ function and it gets called but has fields that are not the defaults.

If I put in similar "if the user is not a betatester then explicitly remove the 'agenda_layout' field" code, it works, but not sure why.

aaronoppenheimer:
Right. Because it's a class variable and a list is mutable, making changes changes all of the 'fields' for all of the instances of the view. But declaring it in an __init__ makes it an instance variable (and overrides the "none" class variable). Works better now!

bklang:
> Right. Because it's a class variable and a list is mutable, making changes to the 'fields' list changes all of the 'fields' for all of the instances of the view including future instantiations. But declaring it in an **init** makes it an instance variable (and overrides the "none" class variable). Works better now!

Of course! Great catch, I wasn't even thinking about class vs. instance variables.

bklang:
Most of our discussion so far has been on the effect of the beta flag and making it work right (which I think will be a great feature to have!) But I also have thoughts on the agenda view changes being proposed here.

My first thought is that every time we add additional flexibility, we're committing to maintaining it forever.  My preference would be to simplify to the best view unless there's a really strong argument to keeping multiple versions. 

From what you propose here, I personally find the the current "Weigh In" view the least useful. It tends to hide the information I want most, which is "what is coming next?"  

This still leaves the questions that I believe inspired your work in the first place:
* What gigs still need my input?
* How can I see just the band gigs that I want to see?

Perhaps it would be less UI work if we used a filter setup instead of discrete views. Imagine that we take the "Single List" view from this PR and add options to filter that list. The filters can be to "Show gigs which have a response/have no response" and "Show gigs for All Bands/Band A/Band B". We can even make those filters sticky so that each time you come to the page, it remembers which filters you had in place. This should simplify the amount of UI code a lot while still giving users flexbility if they want it. Thoughts?

aaronoppenheimer:
I don't love the idea of multiple filters just because it makes things more complicated thaMaybe it can just be one filter - show me everything / show me gigs that need response / band a / band b. If you're only in one band then it's just a toggle between "show me everything" and "just show me gigs that need a response" anyway.

I like the underlying structure of "kinds of gig lists" because there other places that can use it - like the "public view" that was in the old gig-o and someone asked for recently.

I'll try a few things.

aaronoppenheimer:
@bklang pull down the branch 'schedule_filtering' and see what you think. It just gives you radio buttons at the top to select between all gigs, gigs that need a plan, and, if you're in more than one band, a button for each of your bands. It needs some detailing but I think it works well. (there's a migration)

PR #510, opened 2024-07-13
Moving to real permissions system for beta tester

aaronoppenheimer:
I previously added a "is_beta_tester" flag to the member object, mainly because I didn't understand how the django permissions system is supposed to work (partly because there was a problem with the member admin page that hid functionality). This removes the is_beta_tester flag again.

Instead, it uses the "beta_tester" permission that was created on the member object way back but never used! With this in place, we can make a "beta testers" group that adds that permission. Then when people are in that group, templates and other things can see. Also I added a "BETA" flag on the navbar if the user is a beta tester.

HAS MIGRATION


aaronoppenheimer:
On second thought, this isn't really worth it. Adding your own by-object permissions is a valid way to do this, and it's not like we're going to have lots of groups of users anyway.

PR #512, opened 2024-07-14
Schedule filtering

aaronoppenheimer:
adds filtering to agenda page: "all" "needs response" and by band if there's more than one.


aaronoppenheimer:
@bklang OK, I think this is in good shape, and it's a way better solution than what I first did. Thanks for the push in this direction.

aaronoppenheimer:
Added "old layout" mode in case people freak about the new filtering layout - which nobody will but you know.

bklang:
This is fantastic! I love it. So much easier and faster.

agenda/tests.py
```diff
@@ -16,52 +16,50 @@
...
-
-
-        # tests that pass if pagination is set to 10        
-        # # first 'page' of gigs should have 10
-        # response = c.get(f'/plans/noplans/1')
-        # self.assertEqual(response.content.decode('ascii').count("xyzzy"), 10)
 
-        # # second 'page' of gigs should have 9
-        # response = c.get(f'/plans/noplans/2')
-        # self.assertEqual(response.content.decode('ascii').count("xyzzy"), 9)
```
bklang, line 64: Should these tests be fixed or removed?

aaronoppenheimer: They're removed already - not sure why you see them.

bklang:
One more minor thing I found: the `Hide canceled gigs` preference is not respected with the new filtered view. 
Other than that, I think this is good to go. I'm so excited to use it myself.

bklang:
> One more minor thing I found: the `Hide canceled gigs` preference is not respected with the new filtered view.

I just pushed up a fix for this. It only affected the band-specific list of gigs.

aaronoppenheimer:
Great - if you're happy I'm happy. I'm happy to deploy it though I've never done one with a migration. I assume you merge, wait for the deployment action to complete, then log into the server and just run the migration through the shell?

bklang:
Yes, that’s correct. If you want to give it a try, go for it. I’m around to help out in case of trouble.
-- 
Ben Klang
Vice President of Business Technology
Power Home Remodeling 
O: 610.874.5000 <tel:610.874.5000;2242> x4000
PowerHRG.com <http://powerhrg.com/>

PR #520, opened 2024-07-29
Create setting for emailing gig owner when member rsvps

Jessegottlieb:


aaronoppenheimer:
Hi Jesse - very happy to have you contributing!

Thinking about this, we probably don't to send an email back to the gig owner whenever anyone changes their answer, since we're already sending a lot of email, and this could mean we send out twice as much mail for every gig, or even more. I guess huge bands would turn it off out of necessity, but even for a mid-size band, it's a lot of mail.

Is there a specific use case you're thinking about? How about something like a daily digest, or even just a one-week-before-the-gig alert that shows the RSVP status, which would give the leader time to send a reminder?

A

Jessegottlieb:
Hey Aaron, I definitely get where you're coming from about email noise! My
band leader requested this feature since he often loses track of who has
responded and sometimes feels gaslit when someone changes their response to
a gig since the last time he checked gig-o. The emails will serve as a gig
response history and help band leaders have more immediate feedback when
venues need timely responses. My thought was to add a setting that will
default to OFF so that these emails are opt-in.  Your other ideas of a
daily digest or gig reminder sound good, but I don't think they serve my
band leader's request.

I'm happy to code up what I'm thinking of and if you're not into it, I'll
iterate on it or take an entirely different approach. I'm new to Django and
Python so it may take me a while to get a working prototype 🙃

go3/.env.bak
aaronoppenheimer, line 1: I think you probably added this by accident?

aaronoppenheimer:
It's also just about keeping the amount of email down because it's not free. But I see the use case, and definitely give it a shot and see how it works.

Another approach could be to turn on the "simple history" system for gig plans. That's what gigs use to know what has changed when an update email is sent - it just squirrels away previous versions and can do a diff or whatever. I can imagine using that so you could see which plans have changed when looking at the gig page rather than emailing.

Maybe @bklang has an opinion?

aaronoppenheimer:
I am implementing this in a different way - should be PR'd soon.

Jessegottlieb:
Oh great! Sorry I got sidetracked and completely forgot about this. Lmk if
you'd like any help - code reviews or design ideation, etc...

aaronoppenheimer:
Addressed in #619 (currently in beta)

PR #555, opened 2024-10-05
More past gig work

aaronoppenheimer:
Fixed issue with full-day gigs being considered in-the-past too soon.


aaronoppenheimer:
Blah - there's still an issue if it's late in the day east-coast time. I really need to convert everything to the local time of the user and compare dates, and have been avoiding it. It's not straightforward b/c "plans" might be for multiple bands in different timezones. Not usually, but not never. So every plan needs to be compared to the date in the timezone for the band that has the gig. Will take some time to work out.

PR #563, opened 2024-10-12
Skip time validation when unchanged

### LINKED ISSUE #557: In-progress gigs cannot be edited
This is very much an edge case, but I figured it'd be worth documenting/asking if it's worth fixing. No part of a gig page can be edited once the gig's call/start time passes, since it validates that the time must be in the future. I was trying to edit a set list, and couldn't change it unless I also set the call time to a time in the future. Would it be difficult to bypass the "time is in the future" validation if the time hasn't changed?
IMG_3421

bklang:
Fixes #557

### DIFF
--- a/gig/forms.py
+++ b/gig/forms.py
@@ -34,6 +34,9 @@ class GigForm(forms.ModelForm):
             label_suffix='',
             **kwargs
         )
+
+        # Keep track of the initial date so we can see if it changed in the form validation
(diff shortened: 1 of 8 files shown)


aaronoppenheimer:
Wow, such a load of stuff that should have been fixed a long time ago. Plus a fix for the actual issue! Looks good to me.

PR #570, opened 2025-01-11
No more timezones

aaronoppenheimer:
This branch does away with timezones:

- The Gig model now has date/setdate/enddate as no-timezone datetimes, using Ben's trickery to get it to work with the database
- There is no timezone in the band's edit page (though it's still on the Band model for now)
- The various views like calendar and agenda view work. Agenda uses the timezone determined by the GET request to decide whether a gig is in the past.
- Calfeed dates do not have timezones, which is interpreted by (hopefully all) calendar apps as "just use my timezone"
- There are a few data migrations that fix any current gigs to work properly: squirrel away the current timezone-aware datetime and make a new zone-free datetime that is adjusted for the band's timezone.

I have tested this by building a database under the current main branch, creating gigs with various date/times that could give trouble, switching back to this branch and executing the migrations. Seems to work as hoped.


aaronoppenheimer:
So, @bklang, this seems to work, but it's terrifying. The migrations stash the current timezone-aware dates on the gig in case something breaks, but I'll keep testing it.

aaronoppenheimer:
OK, some more confidence. I did this:

- Check out the main branch, load up the contents of the live DB
- For every gig, consider the date, setdate, gigdate - adjust them to local tz using the gig.band.timezone
- Write the date,setdate,gigdate to a file
- Switch to this branch, run the migrations
- Dump the dates in the same way
- Verify that the old dates (adjusted to local) are the same as the new dates (without timezones)

This worked fine - that's good. I would appreciate if you could repeat this and make sure I didn't do anything dumb in the test.

bklang:
I'm testing this now, and so far, so good!

I'm putting particular focus on the calendar feed part. I used a copy of the live database and checked all gigs going back and forward a several months for Seed & Feed and Black Sheep. Confirmed that gigs with and without times specified are showing up correctly.

Next I'm going to enter a bunch of gigs via the UI and see if I can find any edge cases. Also going to test making my browser report various timezones.

aaronoppenheimer:
That's great! I can't wait until the timezone nightmare is over.

PR #582, opened 2025-02-19
Add timezone back into caldav feed

aaronoppenheimer:
It turns out that google calendar (at least) really does need a timezone. As a result, the recent move away from timezones has broken the calfeed - every event is assumed by google to be in UTC. This PR adds back the ability for a band to edit its timezone, and uses that timezone in the calfeed. THis may not be the long-term answer but it should fix the calfeeds for now.

### DIFF
--- a/band/forms.py
+++ b/band/forms.py
@@ -35,7 +35,7 @@ class BandForm(forms.ModelForm):
     class Meta:
         model = Band
         fields = ['name', 'shortname', 'hometown', 'description', 'member_links', 'website',
-            'new_member_message', 'thumbnail_img', 'images', 'default_language',
+            'new_member_message', 'thumbnail_img', 'images', 'default_language', 'timezone',
             'anyone_can_create_gigs', 'anyone_can_manage_gigs', 'share_gigs',
             'send_updates_by_default', 'rss_feed', 'simple_planning', 'plan_feedback']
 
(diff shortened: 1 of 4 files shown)


aaronoppenheimer:
And... I've discovered that there's a bunch of stuff that was added as part of the timezone switch that we don't need, like members having their own timezone. I'll strip that out after this is merged and things are stable again.

aaronoppenheimer:
@bklang when you get a chance, please make sure this will fix the acute issue

aaronoppenheimer:
I figured out what happened - the new timezone plan relies on knowing the users's timezone for producing the agenda page and, as it turns out, the cal feed. But I turned off the member timezone thing at one point and didn't turn it back on, and my tests didn't catch it. So instead of this, I'll fix the member timezone and use it in the calfeed.

bklang:
Thanks @aaronoppenheimer . Just to make sure I'm understanding your latest comment, you want to do more work on this PR prior to merging, right?

bklang:
And I just saw it's back in Draft status - I'll hold off until I hear more from you

bklang:
Looking at this PR, I'm not sure this is what we want. If we add a timezone to the icalendar feed, then we will (mostly) fix Google, but break Apple Calendar.

If we don't have a way to work around Google's icalendar limitations, we may have to go back to the old way of managing timezones. And then make it easier to specify when a particular gig is in a different timezone than the band's normal timezone.

PR #585, opened 2025-02-22
Timezone Updates

aaronoppenheimer:
Adds some fun things:

- Autodetects a user's timezone when they look at the home page
- Properly shows gigs on the agenda page that are in the future (or started but not ended)
- Uses the user's timezone to add timezones to the calfeed
- Fixes some time-validation issues on the gig edit page

This is not a fix for everything - the case where a user changes timezones but doesn't log into the gig-o is still broken since the calfeed will be using the wrong timezone.


aaronoppenheimer:
@bklang have a look!

bklang (approved):
I like that it requires minimal from the user to be correct in most cases. If this works well, we can consider updating the timezone preferences on visits to URLs other than the base agenda route.  Looks good!

PR #590, opened 2025-02-27
Build Gig API

PresidentNick:
Hi, I'm working on my band's website and I would like to create an api to retrieve upcoming gigs. I also have a few more features planned in this [issue](https://github.com/Blowcomotion/blowcomotion.org/issues/33)

In this PR:

- Run `isort` on modified files
- Add `django_ninja` dependency
- Add `api_key` to Band model
- Add `api_key` to band_detail.html
- Generate `api_key` on `Band` save
- Add migration file that generates `api_key` for all `Band` objects
- Add `/api` URL with `BandAPIKey` auth
- Add `/gigs` API route that lists all gig for the authenticated band
- Add automatically generated docs at `/api/docs` URL

TODO: Unit test the `/api/gigs` endpoint

Planned features:
- Add `GET` endpoint filters
- Add `POST` endpoint to create gigs
- Add `PUT` endpoint to edit gigs


aaronoppenheimer:
Hi Nick - this is great. We had a functioning API on the old version of the gig-o, which was built by someone who then made a native phone app using it. There was a start at using GraphQL for the new gig-o, but it didn't get far (and now I'll strip it out in favor of what you're doing).

I'm not familiar with django-ninja so it'll take a little time before I can get my head around it, but at first blush it looks very cool - I take it you've used it before?

A couple of questions/comments:

- Is the intent that every member of a band can have access to the api key for the band? I guess that's fine if the API is read-only, since it's like a calendar feed or RSS feed - but if you're planning to be able to create gigs through the API then we have to respect things like the "any user can create gigs" setting, which seems to mean we need to authenticate by user. We'll also want to be able to log who did what.
  - Would it be better to have an api key assigned per user? Maybe band admins automatically get an api key but non-admins don't? That would let you use the authentication decorators and functionality we've already built to handle requests.
- There are also settings like "hide this gig from the calendar feed" and the like - how can those be respected? I guess if the entire gig is sent over the wire, those per-gig preferences will go with it.
- Not asking you to implement this, but we'll eventually probably want to build this out for the user's personal gig feed (which is what the old app used) so it's worth keeping in the back of your mind.
- The gig has links to 'member' objects, like the gig contact. How will you handle this? Will django-ninja just send over the pk for the member? For this purpose maybe sufficient to just send the name of the contact member over?

Again: thrilled that you're taking this on.

PresidentNick:
Hi @aaronoppenheimer thanks for taking a look at my PR! Oh yeah, I remember seeing that app a few years ago, but I never used it. I just went to the gig-o-matic website on mobile since it works well.

I was up late last night coding my band website (we also use django) and I got the idea for adding the API to gig-o-matic so we can automatically generate content on our website! I was not able to go to sleep until I got my idea started here, haha.

I started by checking the existing graphql implementation, but I realized it would not work for my needs after playing around with it a bit. I have never used graphql or django-ninja; it's my first time. The django-ninja documentation and the layout, readme, and fixtures of the gig-o-matic codebase made it really simple for me to run locally and add the API! Thanks for the excellent documentation!

- Sharing the API key with the whole band is not my intent, I just wanted to get this started and collect feedback before I was finally able to sleep.
  - Those are good ideas! I think we should have:
    - band read key 
      - share with whole band for anyone to create apps
    - band read/write key
      - only display to band admin, band admin can share with trusted apps/engineers
    - user read/write key
      - show all gigs in every band they belong to for this user
      - allow user to create/edit gigs for each band they belong to if "any user can create gigs" setting is true for that band
  - What do you think of these ideas?
- Here is a sample API response:
```
{
    "gigs": [
        {
            "id": 1,
            "title": "tes gig",
            "band_id": 117,
            "details": "do the purple band thing",
            "created_date": "2025-02-27T06:23:36.782Z",
            "last_update": "2025-02-27T06:23:36.782Z",
            "date": "2025-03-04T13:00:00",
            "datenotes": "These are date Notes",
            "safe_date": null,
            "address": "Stargazer",
            "status": 1,
            "email_changes": true,
            "is_archived": false,
            "is_private": false,
            "creator_id": 300,
            "invite_occasionals": true,
            "was_reminded": false,
            "hide_from_calendar": false,
            "default_to_attending": false,
            "trashed_date": null,
            "contact_id": 300,
            "setlist": "moliendo\r\nsoulfinger",
            "setdate": "2025-03-04T13:30:00",
            "enddate": "2025-03-04T14:00:00",
            "safe_setdate": null,
            "safe_enddate": null,
            "is_full_day": false,
            "has_call_time": true,
            "has_set_time": true,
            "has_end_time": true,
            "dress": "purple",
            "paid": "gratis",
            "postgig": "go home",
            "leader_id": null,
            "leader_text": "nick@example.com",
            "plans_locked": false,
            "rss_description": "",
            "cal_feed_id": "75210e4e-1d0e-4eb5-b837-9cf3a879fd85"
        },
        {
            "id": 3,
            "title": "blowco gig 2",
            "band_id": 117,
            "details": "",
            "created_date": "2025-02-27T07:42:22.318Z",
            "last_update": "2025-02-27T07:42:22.318Z",
            "date": "2025-03-07T00:00:00",
            "datenotes": "",
            "safe_date": null,
            "address": "",
            "status": 0,
            "email_changes": true,
            "is_archived": false,
            "is_private": false,
            "creator_id": 300,
            "invite_occasionals": true,
            "was_reminded": false,
            "hide_from_calendar": false,
            "default_to_attending": false,
            "trashed_date": null,
            "contact_id": 300,
            "setlist": "",
            "setdate": null,
            "enddate": null,
            "safe_setdate": null,
            "safe_enddate": null,
            "is_full_day": true,
            "has_call_time": false,
            "has_set_time": false,
            "has_end_time": false,
            "dress": "",
            "paid": "",
            "postgig": "",
            "leader_id": null,
            "leader_text": "",
            "plans_locked": false,
            "rss_description": "",
            "cal_feed_id": "ad7f3e9e-3ced-41d0-bc47-fe7339c0930d"
        }
    ],
    "count": 2
}
```
- Yeah the API returns `hide_from_calendar` and it only displays member PKs so we should resolve that to the member name, good catch!

PresidentNick:
Here's what I got so far!

image

PresidentNick:
@aaronoppenheimer I've only got `api/gig/` so far, but I think that's enough for me to continue building the band website to retrieve upcoming gigs. Can you review and test it locally please to get your feedback?

I will do the `/api/member` `/api/band` APIs in another PR

PresidentNick:
Looks like filters don't work

PresidentNick:
Oh we shouldn't allow getting gigs that don't belong to the authenticated band and also not allow getting gigs for a band the authenticated member doesn't belong to

PresidentNick:
get_gig checks for the auth key first before returning the gig now

aaronoppenheimer:
Hi Nick - I’m out of town for a couple of days but will give it a try when I’m back.

I think determining if it’s a band or member by just looking up the key is cool, but I still think that when it comes to writing, we’ll want to authorize by person. In fact, can’t you do everything with just the user’s key?

Another thing to ponder: bands have “public” calendar feeds, so if an unauthorized person requests a band’s gigs they could just get the public list.

PresidentNick:
Hi Aaron, hope you have a great trip!

Yeah that's true, I hadn't considered we can everything with just the user key. I thought, maybe, what if a member left a band and they had the only key to the website/application, but then I just realized the band admin would assign a new member to maintain the website/app and use their key instead.

Thanks, I didn't know about the public calendar feeds. That's useful to know! I will remove the auth from the endpoint and return the appropriate data based on auth key.

aaronoppenheimer:
Taking a look now - seems pretty good! A few things to ponder:

- I'd like @bklang to weigh in on whether the user's API key is enough security. Since you can't get someone's key unless you are in their account anyway, I suppose it's a similar level of security..? I'm out of my depth and maybe Ben has an opinion
- The version of the gig-o you're building on does not use timezones in the gig's date/calldate/enddate. We used to store timezones with gigs, then took it out a couple of weeks ago in an attempt to simplify thing, but it's about to go back in because it wasn't simpler. So the dates you ship over with the gig will be UTC-timezone'd dates. The band has a timezone that you can use to shift the dates around. Individual users have a "current_timezone" preference, too, which you could use if that makes more sense for what you're building.
- For things like gig status choice - what about language? It might be better to return the choice as the index of the answer, and use the 'plan_status_choices' endpoint with a /[language] parameter to return the strings in whatever language. Caring about it is probably in the far future but better to plan for it than have to retrofit everything later.

PresidentNick:
I tested changing the language for a user and it looks the status choice resolvers for confirmation choices are using translations!

{
  "count": 4,
  "gigs": [
    {
      "id": 3,
      "member_status": "No Plan",
      "band": "Blowcomotion",
      "contact": "Nick",
      "leader": "",
      "date": "2025-03-07",
      "call_time": "00:00",
      "set_time": "",
      "end_time": "",
      "gig_status": "Non confirmé",
      "datenotes": "",
      "title": "blowco gig 2",
      "details": "",
      "setlist": "",
      "dress": "",
      "paid": "",
      "postgig": "",
      "is_full_day": true,
      "address": "",
      "is_archived": false,
      "is_private": false
    },
    {
      "id": 5,
      "member_status": "Je serai là !",
      "band": "Cantankerous Folderol Band",
      "contact": "Amina Priscilla",
      "leader": "",
      "date": "2025-04-02",
      "call_time": "00:27",
      "set_time": "13:30",
      "end_time": "14:00",
      "gig_status": "Non confirmé",
      "datenotes": "",
      "title": "do not show in dmcb",
      "details": "",
      "setlist": "",
      "dress": "",
      "paid": "",
      "postgig": "",
      "is_full_day": false,
      "address": "",
      "is_archived": false,
      "is_private": false
    },
    {
      "id": 1,
      "member_status": "No Plan",
      "band": "Blowcomotion",
      "contact": "Nick",
      "leader": "Nick Pereira",
      "date": "2025-03-04",
      "call_time": "13:00",
      "set_time": "13:30",
      "end_time": "14:00",
      "gig_status": "Confirmed",
      "datenotes": "These are date Notes",
      "title": "test gig",
      "details": "do the purple band thing",
      "setlist": "moliendo\r\nsoulfinger",
      "dress": "purple",
      "paid": "gratis",
      "postgig": "go home",
      "is_full_day": false,
      "address": "Stargazer",
      "is_archived": false,
      "is_private": false
    },
    {
      "id": 4,
      "member_status": "No Plan",
      "band": "Blowcomotion",
      "contact": "Nick",
      "leader": "",
      "date": "2025-03-07",
      "call_time": "00:00",
      "set_time": "",
      "end_time": "00:00",
      "gig_status": "Non confirmé",
      "datenotes": "",
      "title": "Copy of blowco gig 2",
      "details": "",
      "setlist": "",
      "dress": "",
      "paid": "",
      "postgig": "",
      "is_full_day": true,
      "address": "",
      "is_archived": false,
      "is_private": false
    }
  ]
}

aaronoppenheimer:
Huh - why are some of them in English and some not? That's weird..? And I'm not sure how the API would know what language to respond in, since the language is set in the session, and the API doesn't do that. It could, of course - maybe during the authentication function. So it's strange that they're translated at all.

But I also mean - if the *app* that is pulling the information wants to present the information in a different language, it will need a way to get all of the translated strings (or at least the ones for things like this). Maybe it's not worth worrying about now. But I do wonder why it's translating anything now.

gig/api.py
```diff
@@ -0,0 +1,158 @@
...
+
+    Key type is determined by the API key used for authentication and displayed for informational purposes.
+    """
+    api_key = request.auth.get("key")
+    member = Member.objects.filter(api_key=api_key).first()
+    if not member:
+        return JsonResponse({"message": "Unauthorized"}, status=401)
+    else:
+        # get all gigs for the member
+        plans = Plan.objects.filter(assoc__member=member)
```
aaronoppenheimer, line 118: I think you should be able to do all the plan/gig filtering in one shot, with something like

```
p=Plan.objects.filter(assoc__member=member)
if filters.plan_status:
  p=p.filter(status=filters.plan_status)
if filters.gig_status:
  p=p.filter(gig__status=filters.gig_status)
gigs=p.values_list('gig',flat=True)
```
That turns into (I think) just one call to the db.

aaronoppenheimer:
> Huh - why are some of them in English and some not? That's weird..? And I'm not sure how the API would know what language to respond in, since the language is set in the session, and the API doesn't do that. It could, of course - maybe during the authentication function. So it's strange that they're translated at all.

Well, it looks like there's no French translation in the system for "No Plan" which is why it appears in English. But I'm still mystified about how the api knows that it should be using French. The language is stored in a cookie - is that being sent over somehow?

PresidentNick:
It's getting the cookie from the browser since I was using the ninja docs interface. It probably wouldn't send the language cookie on postman or curl. I didn't try it there.

aaronoppenheimer:
Ah, there you go. I was using postman.

PresidentNick:
I'll add a language param to the apj

PresidentNick:
> Huh - why are some of them in English and some not? That's weird..? And I'm not sure how the API would know what language to respond in, since the language is set in the session, and the API doesn't do that. It could, of course - maybe during the authentication function. So it's strange that they're translated at all.
> 
> But I also mean - if the *app* that is pulling the information wants to present the information in a different language, it will need a way to get all of the translated strings (or at least the ones for things like this). Maybe it's not worth worrying about now. But I do wonder why it's translating anything now.

Yeah it could either be a translate API to translate  individual phrases or the whole API is translated with a parameter. But we'd only translate the values, not the keys.

gig/api.py
PresidentNick: thanks I'll try that later!

aaronoppenheimer: By the way - don't feel compelled to change it, we're not that big into super-efficiency or something. I'm the worst, in fact.

bklang:
Hey @PresidentNick thanks for your efforts here! So many cool possibilities!

I haven't had a chance to review the implementation yet, but I did want to weigh in on the translations thing:

> > But I also mean - if the _app_ that is pulling the information wants to present the information in a different language, it will need a way to get all of the translated strings (or at least the ones for things like this). Maybe it's not worth worrying about now. But I do wonder why it's translating anything now.
> 
> Yeah it could either be a translate API to translate individual phrases or the whole API is translated with a parameter. But we'd only translate the values, not the keys.

I think it's important for the API to focus on the job of exchanging data, and not worry about how that data is displayed. Translations are a concern of the presentation layer. This distinction is important because it's really impossible for the API to anticipate all the ways in which the data it serves will be used.

For example, having gig status *always* be one of `['Unconfirmed', 'Confirmed', 'Canceled', 'Asking']` means that the API consumer (your website) could reliably make decisions like, for example, what kind of icon to show. If the field were translated, that job would be much harder. For this reason, I'd prefer to see all fields use normalized values, not translated ones. English values are what the site uses internally, and are probably the most logical choice.

member/models.py
```diff
@@ -194,6 +200,7 @@ def as_email_recipient(self):
     def save(self, *args, **kwargs):
         """ when creating members with a form, the usermanager isn't used so we have to override save """
         self.email = self.email.lower()
+        self.api_key = self.api_key or str(uuid.uuid4())
```
bklang, line 203: It looks like this allows users to specify their own API key, with a default value being a UUID. UUIDs are ok as API keys, but not great. While it's highly impractical that an attacker could guess them, there are better options for secure key generation.

I do think key generation is the way to go. Users providing their own API key are pretty likely to pick an insecure value.

Finally, I'd like API key generation to be an explicit step. Most users won't need an API key, so by not setting one, we reduce any risk of leak.

For generating the API key, I suggest using the Python [secrets library](https://docs.python.org/3/library/secrets.html)

```Python
import secrets
secrets.token_urlsafe(30)
```

This will return a 30-character API key that is also URL safe. A pretty good balance between security and convenience.

gig/api.py
```diff
@@ -0,0 +1,158 @@
...
+    def resolve_member_status(obj, context):
+        plans = obj.member_plans
+        if plans:
+            auth_key = context.get("request").auth.get("key")
+            member = Member.objects.filter(api_key=auth_key).first()
+            if member:
+                plan = plans.filter(assoc__member=member).first()
+                if plan:
+                    return next((str(choice[1]) for choice in PlanStatusChoices.choices if choice[0] == plan.status), "Unconfirmed")
+        return "Unconfirmed"
```
bklang, line 57: If I'm reading this correctly, any band member other than the user whose API key is in use will always show "Unconfirmed". Is this really what you want? I would think an authenticated request would be able to see the actual plans for all band members, just like they would in their web browser.

bklang:
> * send back 1 for "Definitely," 2 for "Probably," etc. and let the front-end decide how to display that. An option is to have another api to translate a 1 into a word in English or French or whatever.

You're right that we have an internal enum that maps those integers to the code words "Definitely", "Probably", etc. However, personally I don't like using integers in external APIs because it makes them harder to read. You have to document the integer values for API consumers, vs. just reading the JSON.

If we ever had a need to change "Probably" in the UI, we could do that by making use of the translation layer. Even in English, the Gig-O website could choose to show the code value `Probably` as "Pretty Likely" if we added that translation to the English file. We actually do this for the gig status "Confirmed" which shows up as "Confirmed!" in the UI thanks to the translation*. So, in practice, the `Probably` value is equivalent to a number in stability, but much more readable.

If we want to support both status codes and help out API consumers by providing translated values, we could do that. Just as two separate fields:

```JSON
{
  "gig_status": "Confirmed"
  "gig_status_localized": "Confirmé"
}
```

Personally, I think this is a bit overkill as any app which uses our API and needs localized strings will have its own translation database. But I'm not entirely against being helpful in this way if you guys want it.

* This is a small lie - there's some funky code that pre-translates it to "Confirmed!" - but the underlying point is true that the enum uses a fixed code word and the UI shows a translated string.

aaronoppenheimer:
> You're right that we have an internal enum that maps those integers to the code words "Definitely", "Probably", etc. However, personally I don't like using integers in external APIs because it makes them harder to read. You have to document the integer values for API consumers, vs. just reading the JSON.

Yeah, I see your point. At that point, the text is just a token like the number. Since nobody actually planning on changing any (English) text any time soon, it's probably better to be readable.

PresidentNick:
> 
> (I mention "old gig-o" and "new gig-o" - in case you weren't around for it, we replaced the gig-o with a brand-new version last Spring.)
> 

Yeah, I remember things being unsure in my band what we would do if gig-o-matic would shut down, but I'm glad y'all could keep it going!

member/models.py
PresidentNick: @bklang I wasn't intending to let users define their own API key, the intent was to generate the key if it didn't already exist. I didn't consider the Member init function where they could pass in a key they defined before saving the member. I can refactor this so it doesn't allow that.

gig/api.py
PresidentNick: @bklang  No, I was just leaving a default value, but I can refactor this to throw an error if it never finds the gig or plan instead of lying about the status.

PresidentNick:
@bklang good idea I like this!

```
{
  "gig_status": "Confirmed"
  "gig_status_localized": "Confirmé"
}
```

Should it also return the integer value for the status?

PresidentNick:
> > You're right that we have an internal enum that maps those integers to the code words "Definitely", "Probably", etc. However, personally I don't like using integers in external APIs because it makes them harder to read. You have to document the integer values for API consumers, vs. just reading the JSON.
> 
> Yeah, I see your point. At that point, the text is just a token like the number. Since nobody actually planning on changing any (English) text any time soon, it's probably better to be readable.

oh so should it just return the resolved status in English?

gig/api.py
PresidentNick: I did a little research and I found this in the docs:

django-ninja supports [returning querysets](https://django-ninja.dev/guides/response/#returning-querysets) and django doesn't [evaluate](https://docs.djangoproject.com/en/5.1/topics/db/queries/#querysets-are-lazy) stacked querysets or [chains](https://docs.djangoproject.com/en/5.1/topics/db/queries/#chaining-filters) of querysets until we return the queryset in the  API

member/tests.py
```diff
@@ -1300,3 +1300,17 @@ def test_member_detail_access(self):
...
+        self.client = Client()
+        self.member = Member.objects.create_user(
+            email='member@example.com', password='password123'
+        )
+
+    def test_generate_api_view(self):
+        self.assertFalse(self.member.api_key)
+        self.client.force_login(self.member)
+        self.client.get(reverse('member-generate-api-key'))
+        self.assertTrue(self.member.api_key)
```
PresidentNick, line 1316: The view seems to work in the browser, I am able to generate an `api_key` when I press the button in `member_detail.html` but in this test, this assertion fails, but I haven't determined why yet.

Any ideas here? @aaronoppenheimer @bklang?

aaronoppenheimer:
> > > You're right that we have an internal enum that maps those integers to the code words "Definitely", "Probably", etc. However, personally I don't like using integers in external APIs because it makes them harder to read. You have to document the integer values for API consumers, vs. just reading the JSON.
> > 
> > 
> > Yeah, I see your point. At that point, the text is just a token like the number. Since nobody actually planning on changing any (English) text any time soon, it's probably better to be readable.
> 
> oh so should it just return the resolved status in English?

I agree with Ben that the return values should be meaningful instead of just numbers. If the API returns the in-English status value, this is accomplished. The API user can then use that to decide what to display. If the application wants to display in a language other than English, it can do its own translation (or there can be some way to get translations for common stuff through the API, I guess, but that can be figured out in the future). We'll just have to be sure that the English values we use in the "choices" objects don't change because they'll break apps using the API, but that's fine - we should probably decouple the values in the "choices" and the actual display text in the gig-o anyway.

bklang:
> > > > You're right that we have an internal enum that maps those integers to the code words "Definitely", "Probably", etc. However, personally I don't like using integers in external APIs because it makes them harder to read. You have to document the integer values for API consumers, vs. just reading the JSON.
> > > 
> > > 
> > > Yeah, I see your point. At that point, the text is just a token like the number. Since nobody actually planning on changing any (English) text any time soon, it's probably better to be readable.
> > 
> > 
> > oh so should it just return the resolved status in English?
> 
> I agree with Ben that the return values should be meaningful instead of just numbers. If the API returns the in-English status value, this is accomplished. The API user can then use that to decide what to display. If the application wants to display in a language other than English, it can do its own translation (or there can be some way to get translations for common stuff through the API, I guess, but that can be figured out in the future). We'll just have to be sure that the English values we use in the "choices" objects don't change because they'll break apps using the API, but that's fine - we should probably decouple the values in the "choices" and the actual display text in the gig-o anyway.

For a good breakdown of the various statuses, see https://github.com/Gig-o-Matic/GO3/blob/main/gig/util.py and https://github.com/Gig-o-Matic/GO3/blob/main/band/util.py

member/tests.py
PresidentNick: that was it, thanks!

PresidentNick:
> 
> For a good breakdown of the various statuses, see https://github.com/Gig-o-Matic/GO3/blob/main/gig/util.py and https://github.com/Gig-o-Matic/GO3/blob/main/band/util.py

Yeah! I used those for `/api/gig_status_choices` and `/api/plan_status_choices`

templates/member/member_detail.html
```diff
@@ -71,6 +71,19 @@
...
                         </div>
                     </div>
+                    <div class="row mt-4">
+                        <div class="col-sm-3 col-12">
+                            <p>API Key</p>
+                        </div>
+                        <div class="col-sm-9 col-12">
+                            {% if member.api_key %}
+                                <p>{{ member.api_key }}</p>
+                                <p><a class="btn btn-primary btn-sm" href="{% url 'member-revoke-api-key' %}">{% trans "Revoke API Key" %}</a></p>
```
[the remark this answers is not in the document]
aaronoppenheimer: So easy:

```
from django.utils import translation
translation.activate(member.preferences.language)
```
Then:
```
>>> PlanStatusChoices.DEFINITELY.label
'Definitely'
>>> translation.activate('de')
>>> PlanStatusChoices.DEFINITELY.label
'Sicher'

```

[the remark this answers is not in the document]
PresidentNick: neat! what about in the front end with the `trans` template tag? Will this be sufficient, do I need to create a translation file entry?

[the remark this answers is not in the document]
aaronoppenheimer: It’s all written up here: https://github.com/Gig-o-Matic/GO3/wiki/Internationalization

aaronoppenheimer:
I'll keep looking at this, but I don't think we want to merge it into main until after @bklang and I finish the timezone-related fixes. They probably don't overlap, it just seems like two big things to deal with and since the timezone thing is halfway there, I'd rather wait on the api.

PresidentNick:
@aaronoppenheimer thanks for letting me know! As far as I can tell, the overlap is that both PRs add member migrations, but that can be be fixed after one of them is merged by remaking the migration.

Thanks for fixing the timezones! It was making things a little confusing for showing up at gigs.

templates/member/member_detail.html
aaronoppenheimer, line 81: Maybe put the 'revoke' button on the same line as the key, like the email 'test' button above does.

PresidentNick: image

Hmm, do you think this looks cramped?

image

definitely on mobile!

member/views.py
```diff
@@ -63,6 +71,24 @@ def verify_requestor_is_in_user_band(request, user):
     return True
 
 
+@login_required
```
aaronoppenheimer, line 74: Since these are not really 'views' it'd be better to put them in the helpers file. (I haven't been consistent with this - the 'delete_invite' view isn't really a view either.)

templates/member/member_detail.html
aaronoppenheimer: Also, consider putting the API key in its own panel at the bottom of the page, below preferences. 99% of the users (at least) will never click this and since it's jargon-y, it ought to be less prominent.

aaronoppenheimer:
This looks good to me. I've merged in the timezone fix; let us give it a day or so to make sure nothing catastrophic happened before merging this in so we don't have to back it out.

bklang:
I'd also like one more chance to play with this before merging. Most of my time recently has been spent on the timezone fixes. I'll commit to getting that done by Tuesday with either approval or comments.

bklang:
@PresidentNick We've had some pretty big changes on the `main` branch, would you merge that in so we can test the combination?

PresidentNick:
@bklang git says my branch is up to date, I merged it a few days ago [here](https://github.com/Gig-o-Matic/GO3/pull/590/commits/6516652329996732d87d4b19418aad34201da708)

bklang:
> @bklang git says my branch is up to date, I merged it a few days ago [here](https://github.com/Gig-o-Matic/GO3/pull/590/commits/6516652329996732d87d4b19418aad34201da708)

Oh, I missed that. Thanks for being on top of it!

PresidentNick:
I tested this locally with my band website, and it works!!! woooooooohoooooooo!!

[check it out here!](https://github.com/Blowcomotion/blowcomotion.org/issues/33#issuecomment-2716530512)

gig/api.py
aaronoppenheimer: Another issue - only *confirmed* members should be able to see gigs for a band. So the assoc has to have the CONFIRMED status.

gig/api.py
```diff
@@ -0,0 +1,163 @@
...
+    return [{"id": status[0], "name": status[1]} for status in GigStatusChoices.choices]
+
+@router.get("/{gig_id}", response={200: GigSchema, 404: Message})
+def get_gig(request, gig_id: int):
+    """Retrieve a specific gig by its ID if it belongs to the authenticated member plans."""
+    api_key = request.auth.get("key")
+    member = Member.objects.filter(api_key=api_key).first()
+
+    if member:
+        gig = Gig.objects.filter(pk=gig_id).first()
```
aaronoppenheimer, line 157: I think this should be checking to make sure the member is in the band and is a confirmed member.

gig/api.py
PresidentNick: Good catch! I'll add that in

gig/api.py
PresidentNick: @aaronoppenheimer it's in now!

PresidentNick: Added tests for that too in `list_all_gigs` and `get_gig`

gig/api.py
```diff
@@ -0,0 +1,164 @@
...
+    def resolve_leader(obj):
+        return obj.leader.name if obj.leader else obj.leader_text
+    
+    @staticmethod
+    def resolve_date(obj):
+        return obj.date.strftime("%Y-%m-%d") if obj.date else obj.setdate.strftime("%Y-%m-%d")
+    
+    @staticmethod
+    def resolve_call_time(obj):
+        return obj.date.strftime("%H:%M") if obj.date else obj.setdate.strftime("%H:%M")
```
bklang, line 87: Are these both meant to point to the same date field? I see 4 date methods here, but Gigs only have 3 dates:
* `gig.date` which is the call time
* `gig.setdate` which is the start time
* `gig.enddate` which is when the gig is done

bklang: And right after I posted that, I realized that one was date and one was time - you can ignore me

bklang (approved):
Ok, I've had a chance to play with it, and it looks good to me! Stamping my approval on it. I'd like to also give @aaronoppenheimer a chance to give his formal approval too before merging.

Two suggestions for future work:
1. In the UI where the user can generate an API key, it would be great to have a link to the API documentation. I love that we have documentation! Making it more accessible will make this API even more useful to future developers.
2. I think the `whoami` endpoint could provide more useful data, and removing the API key from the return value reduces the risk of leaking sensitive information.

Neither of these are blockers though. Onward!

PresidentNick:
@bklang Thank you for the review!

1. Added the url link in `member_details.html` and named the docs now!
image
2. Yeah I'll probably delete that endpoint in favor of the `/api/member` endpoint in the future to get information about the member

PresidentNick:
another thing for the future is to display public gigs in `list_all_gigs` and `get_gig` when accessing the API without auth key

bklang:
> * I'd like @bklang to weigh in on whether the user's API key is enough security. Since you can't get someone's key unless you are in their account anyway, I suppose it's a similar level of security..? I'm out of my depth and maybe Ben has an opinion

I don't think I ever responded to this.

I think sticking to a per-user API key is fine for now, and I'm mostly comfortable with how it is currently implemented. The only remaining concern I have with the API key is that it is stored unencrypted in the database. Ideally it would be hashed, much as we do for passwords. That way a leak of the database does not constitute a leak of any credentials.

Before we go adding any additional authentication methods, I think we should discuss further. I'm not excited about having a band API key, for example. And I would obviously want to take a much closer look if we allow the API to start making writes to the site.

PresidentNick:
also we must account for timezones in a future API release

> The version of the gig-o you're building on does not use timezones in the gig's date/calldate/enddate. We used to store timezones with gigs, then took it out a couple of weeks ago in an attempt to simplify thing, but it's about to go back in because it wasn't simpler. So the dates you ship over with the gig will be UTC-timezone'd dates. The band has a timezone that you can use to shift the dates around. Individual users have a "current_timezone" preference, too, which you could use if that makes more sense for what you're building.

PresidentNick:
@bklang do you have any recommended methods to define an encrypted API key field in the member model?

aaronoppenheimer:
> also we must account for timezones in a future API release

I'm traveling and can't test it - does the API just return the gig times in UTC?

aaronoppenheimer:
Other than the enhancements we're talking about, it looks good to me!

One thing we might consider - you could make the API key available only to users with the "is_beta_tester" flag set. In the template, just put {% if user.is_beta_tester %} ... {% endif %} around the API key panel. That will let you play with it more without exposing it to everyone just yet. Especially if we think the format of the key might be changing sometime soon, might want to do that.

PresidentNick:
@aaronoppenheimer oh nice, good call on adding the beta tester flag. Currently, it returns the UTC time for the gig

aaronoppenheimer:
I think this is ready to play with, so I'll merge it and do the migration on the live site - woo!

Since the gig returns the date/time in UTC, though, the real use seems limited to testing for now. There are a few ways I can think of handling this, but the simplest is for the gig API to just include the band's timezone - then it's up to the client to do the adjustment to the right zone. Eventually, gigs will be able to have their own timezone stored with them (so I can schedule a gig at HonkTX using Texas local time even when I'm in Boston) and the api can return that instead of the band's timezone.

As you build out the client you're working on, you may have other thoughts about it - let's figure out what'll work best before taking this out of beta.

Great work!

PR #593, opened 2025-03-02
Return of Time Zones

aaronoppenheimer:
Ok! This adds timezone-awareness back to the gig-o. It was actually not a bad exercise to strip out all the timezone stuff that was there originally, then add it back in from scratch - it's much more consistent now.

There's a data migration that goes through every gig and "localizes" the date/setdate/enddate to the band's timezone. We saved the tz-aware dates before as "safe" fields, but since they might have changed, it seemed safer to just re-shift them to the right zone.

I tested it this way:
- reverted to the last commit before we stripped out tz's, and loaded the db with a copy of the live db from then
- dumped all the gig dates into a csv
- switched to the current main and migrated (which strips out the timezones from the gigs in the db)
- switch to this branch and migrated (which adds the tzs back in)
- dumped all the gig dates into another csv
- compared the before/after csvs

I also did a check to make sure that the "safe_date" and the newly-shifted date were the same. So it all seems to work... but would be good to have another set of eyes...


bklang:
I'm putting some eyes on it this morning. Thanks @aaronoppenheimer for taking this on, again!

PR #614, opened 2025-03-22
turned off compressing static files for now

aaronoppenheimer:


aaronoppenheimer:
@PresidentNick For now, just falling back to using a css file with different name and keeping the django-compress thing turned off. It's cool but we need to figure out why it's not working...

PresidentNick:
Ah gotcha, I'll investigate

PresidentNick:
Does the production server serve the static files or is it another server?

https://django-compressor.readthedocs.io/en/stable/scenarios.html

aaronoppenheimer:
ah: https://medium.com/technolingo/fastest-static-files-served-django-compressor-whitenoise-aws-cloudfront-ef777849090c

> Isn’t Django Compressor Incompatible with Whitenoise?
> 
> Normally, yes. But you could use Django Compressor’s offline compression mode to make it compatible with Whitenoise.

and

> 
> Django Compressor
> For performance and security reasons WhiteNoise does not check for new files after startup (unless using Django DEBUG mode). As such, all static files must be generated in advance. If you’re using Django Compressor, this can be performed using its [offline compression](https://django-compressor.readthedocs.io/en/stable/usage.html#offline-compression) feature.

aaronoppenheimer:
Ah. So, we use whitenoise in the gigo, which does the append-a-hash-to-the-filename thing automatically. But - it doesn't work in debug, only in production! That's why I had to force a refresh of my browser when testing. It would have worked on the real site without including the django-compressor stuff. I will back this out!

PR #615, opened 2025-03-22
Widen location column if assoc is not multisectional

PresidentNick:
Once it was deployed I noticed all the extra space in my band since I'm not multisectional. I was developing with multisectional turned on and failed to notice all the extra space available when the user is not multisectional. So now the location is one column wider if not multisectional like this

image


aaronoppenheimer:
I don't think we should do this - it looks better with 'response' buttons lined up; I'd rather that then use up all possible available space.

PresidentNick:
How about this? It uses more of the available space and the buttons stayed lined up

image

aaronoppenheimer:
Yes, I think that's much cleaner - thanks Nick! Is this good to go?

PresidentNick:
Good to go!!

aaronoppenheimer:
Done - great job!

Issue #569, opened 2025-01-08
More detail in calfeed?


aaronoppenheimer:
User writes: "Anyway, the value in my usage is, I'd like to see gigs I've committed to as well as gigs I haven't yet committed to, but be able to differentiate between them. Another band I play with just uses google calendar directly, So with a committed/uncommitted differentiation I can easily decide which I'll commit to when looking at a unified calendar. Showing gigs I've declined is just an extra gravy feature I might not use. I could see scenarios where a band organizer might want gigs on the calendar they're not participating in though."

PR #644, opened 2025-06-21
improve layout of "Preferences" in Member detail page

uneasyalliance:
Change "Preferences" card in member_detail.html to have 2 equal width columns on larger screens. Left column would bunch up into multiple lines. Small screen widths should behave as before (tested by reducing browser width).

Note 1: this is based on "spellfix" branch, which has a separate pull request. Let me know if you prefer this not based on that.
Note 2: I've started working on issue #569 "more detail in calfeed" which will probably add a field to member_detail.html.


uneasyalliance:
Ahoy! Thanks for getting back to me.
Perfectly clear, yeah I think that would clean the "card" up nicely. I think that's the right term for that UI element. I'll check out what's possible with that card layout and see what I can come up with.

uneasyalliance:
@aaronoppenheimer Take a look at this now. Apparently these weird Bootstrap columns can be specified as divisions of 12. So a blank "column" of 1/12 width serves ok as an indent.

aaronoppenheimer:
Ok! The preferences section looks really good, I think. But - the "Bands" card now looks weird with everything crowded on the right side. Maybe the answer there is for the Bands section to mimic what you did in the preferences section - with the band name as the "heading" and then the options aligning with the "label" and "value" columns..?

uneasyalliance:
I just now saw after looking at this again there's effectively 3 columns (but it's not directly coded like that, it's a row within a row), So yeah that's a good idea. The "pending" band membership I had in the example data doesn't show this.

uneasyalliance:
Updated bands card, looks much better.

uneasyalliance:
I was hoping you wouldn't notice they were slightly different :) I guess that won't be bad to tweak Bands to use the layout of Preferences. I suppose Watched Gigs could use the same layout, but right now its (2 (blank), 6, 4) so changing to (1 (blank), 5, 6) might squash the gig detail a bit. In that case I suppose we could swap those two columns if it looked better. The Info tab doesn't have any natural categories, and the labels (left side) are small vs the data (small to large), so I think we should just breathe, and accept the difference.

aaronoppenheimer:
Ha - nothing gets past the all-seeing eyes of the gig-o-matic overlords. Or whatever. I'll be gone for a few days so won't get to this until early next week. Thanks for your hard work!

uneasyalliance:
Feels good to contribute to something I actually use, catch you later

aaronoppenheimer:
OK, I took a look and I think it's using the right set of column definitions now. I'll cop to not having a full understanding of flex and all that stuff, but I think a) the columns align now; b) when you shrink the window to phone size it degrades to a single column. See what you think!

aaron-sao:
Yes, agreed. The preferences panel degrades nicely because (I think) it's using the d-sm-inline class.

aaronoppenheimer:
OK! Check it out and make sure it's right now - thanks!

PR #654, opened 2025-08-06
Calfeed

uneasyalliance:
NOTE this branch is based on member_detail branch, because the following member specific calfeed changes will touch the member_detail page.

Refactor calfeed code to separate band/member calfeed logic. This is a precursor to upcoming changes specific to each type of calfeed. This doesn't really change the behavior, except the member calfeed creation iterates over plans rather than gigs so that user responses will be taken into account. Related tests updated and improved.


uneasyalliance:
I'd say it's ok to merge now. and in fact advantageous to allow the upcoming calendar feed changes for band / member to be done independently. I based this on calfeed branch so that needs to be merged first (or merging this will merge both sets of changes). I believe that's good to go also.

I don't have any feel for the usefulness of the RSS feed, but I'll keep that in mind as I look things over.

aaron-sao:
OK, will merge in. Re: RSS, people ask for it and it's easy to implement. I'll just need the same sorts of things calendar events need: list of events, name of event, details of event. Seems useful to draw on the same code to generate those.

Issue #648, opened 2025-07-16
Calendar event captions


mmendelson222:
I'm working a little bit with the calendar feed (since there doesn't seem to be an active RSS feed).  

Would the superusers be amenable to dropping the "Confirmed" and the band name from the titles?  The way the feed functions is that only confirmed gigs show up there anyway, and because the feed is for just one band, the name of the band is redundant.  

Let me know - I'm happy to do a code pull and do the ahem, heavy lifting. 

<img width="604" height="426" alt="Image" src="https://github.com/user-attachments/assets/4bdaad9e-ff10-4138-9caa-244d397b2180" />

aaronoppenheimer:
It's not as pretty as you'd like, but not hard. The code is in lib/caldav.py. Both feeds use the same code to generate the feed but there's a "is this for a band" flag passed in, and you can use that to adjust the title of the event to suit. The band or member is also passed in, so if you're ambitious, for members you can check the member's preferences and leave out the 'confirmed' if they only want to see confirmed gigs.

Actually, there's no need for the 'is a band' flag since you can check the type of the 'source' and see if it's a band or a member. But if you don't want to clean that up, I don't blame you!

mmendelson222:
Well @uneasyalliance this thread has only been here for two weeks :) 

This is really just a small display issue that solves a simple problem.  Perhaps a different feed url would best suit the issue?  Then either could be subscribed to.  I'm probably not going to have time to look at this in the near future. 

I'm not sure of the best way to get noticed.  I still have #638 pending...

aaronoppenheimer:
Hi both - thanks for working on this stuff. What does "a user option for availability response in the member feed" mean? To be able to say you can be there from within the calendar?

aaronoppenheimer:
Sorry, now I'm confused. Is this what we're talking about?

Band Feed
- Drop "confirmed" and the name of the band from the title of events

Member Feed
- Drop the band name from the title of events if the member is in only one band
- Drop the gig status if the member has checked the 'only show confirmed gigs' preference
- Show the member's availability if they have checked the 'show my answer' preference (not sure what it should be called)

One thing I note is that we shove everything into the title of gigs in the feed (actually the SUMMARY of the caldav event) but there's also a DESCRIPTION which we don't use. Maybe more detail could go in there?

uneasyalliance:
Maybe we're still kicking around ideas?

[...]

Issue #529, opened 2024-08-22
More detail possible in calendar feed?


uneasyalliance:
This is a feature I'd really like to have. I'd like to be able to see in my external (e.g. google or apple) calendars all upcoming gigs, and have some way to visually differentiate my responded/didn't respond,  attending/not attending status. I'm in one band using GigO, and another not, so having a combined local calendar with current status will be very helpful. I haven't fully investigated the .ics format to see where that could go, but I did check that the data is not available currently in the calendar feed.

I considered just screenscraping it, but if you consider this a useful option for users, I'd be willing to do the work to add it. Another option might be a way to download the complete schedule with all details, but I don't know if that's an option where GigO is hosted. I'd also be willing to work on some other pieces. I'm an experienced Python programmer. Let me know your thoughts!

John

aaronoppenheimer:
Hi John - thanks for the input. There is a "Color" property in the iCalendar spec - I've never noticed it before. Apparently it allows clients to override the calendar default color if they want to. I don't know what that would look like but we can experiment and find out. The gig-o has a per-band color setting (which is used by the internal calendar view as well as the default 'schedule' view) - I can imagine giving the user the option to color calendar-feed events by band or by status (yes/no/etc). You can also already tell it not to include "no" events in your calfeed, so maybe that's less useful? How would you propose to use color?

uneasyalliance:
Hi Aaron, thanks for the reply. I was thinking of either a data field (if available in the calendar format), or a subtle title modification, which gives the user's reply status. Whether that's a color, or a bit of text in the title I don't know what the best choice would be. The other band I'm in uses calendar invites, so they are differentiated by outline/solid/hidden for not accepted/yes/no respectively. Something like that would be a big help for my scheduling, but I'm not sure what the capabilities of goog calendar are in terms of custom coloring, it might just be a color per calendar source. In that case some discreet bit of data would do the job, then I could have a process to watch my GigO calendar feed, and then massage it to look the way I want in goog's calendar, then load it in.

uneasyalliance:
I see "(Confirmed)" or "(Uncomfirmed)" for each item in the calendar feed, so perhaps something like that. A user could opt in or out maybe.

aaronoppenheimer:
Huh - I wonder if the gig-o could just send calendar invites.

aaronoppenheimer:
Also https://icalendar.org/iCalendar-RFC-5545/3-2-12-participation-status.html

[...]

PR #671, opened 2025-09-12
update summary and description for band and member calfeeds

uneasyalliance:
Here's the changes previously discussed related to band/member calfeed summary and description.
Take a look and see if this agrees with your thoughts. Still remaining to do is the logic and the user preferences to conditionally show member response and gig status.

band:
  summary:
    gig.title
  description:
    gig.public_description

member:
  summary:
    gig.title
    gig.band.name
  description:
    gig.status
    gig.details
    gig.setlist

I think this closes issue #648
and does part of #529


lib/caldav.py
```diff
@@ -63,20 +63,27 @@ def _make_calfeed_metadata(the_source):
...
+                parts.append(gig.setlist)
+        return "\n\n".join(parts)
 
     event = Event()
     event.add('dtstamp', timezone.now())
     event.add('uid', gig.cal_feed_id)
-    summary = f'{gig.title} ({GigStatusChoices(gig.status).label}) - {gig.band.name}'
+
+    summary = gig.title
+    if not is_for_band: # TODO will be a user option to add band name
```
aaronoppenheimer: This is a change in the current behavior, right? There's currently no option to add the band name (or not). So this will be a change in behavior - I think we should keep the behavior the same until the option is available.

uneasyalliance: The band name is still present for a user calendar (next line cut off in your diff). Gig status is no longer in the summary, and appears only in the description for user calendar. So I think it's the minimal change to get the calendar summary/description close to what was discussed without the user options (yet).

uneasyalliance: I suppose I should rebase this onto the current main....

uneasyalliance: rebased and re-tested, looks good.

PR #684, opened 2025-11-01
Show member attendance history

bklang:
Add a view allowing band admins to view attendance history of individual members. This is intended to help band managers maintain their active member rosters.

A new "Attendance" button is added to the "Member Admin" section of the band page:

Screenshot 2025-11-01 at 7 34 36 PM

The page shows the attendance history for the member:
Screenshot 2025-11-01 at 7 35 50 PM


bklang:
> * Not your fault - it's in the main branch - but the band_detail.html template is missing a closing  for part of the first "card" panel, so the cards aren't separated. No idea how long it's been like that, but maybe fix it as part of this? It's line 95 in the template!

Fixed - thanks!

PR #696, opened 2025-12-08
Add RSVP date handling to GigForm and related templates, resolves #695

### LINKED ISSUE #695: Feature request - It would be helpful to have a dedicated date field for "RSVP By" for the Gigo Event
When we set up a new event in Gigo, we often use it to make a decision if we have enough band members to make the event happen. It would be helpful to have a dedicated, optional, 'RSVP By' date field to provide a clear deadline on when we require folks to have updated their Gigo status. The deadline for needing to know band status could be several weeks before the proposed event date.

We can certainly use the 'details' field or 'Date notes' field to specify that, but, it would be good to have a dedicated field to remind the event creator to set that when needed.

PresidentNick:
- Implemented RSVP date parsing and validation in GigForm.
- Added 'rsvp_by_date' field to Gig model.
- Updated gig_detail.html to display RSVP deadline.
- Enhanced gig_form.html to include RSVP date input with datetime picker.


PresidentNick:
Our band would've found this useful a couple weeks ago but I hadn't considered this idea, so when I saw it pop up I implemented it!

aaronoppenheimer:
In playing with it - I wonder if we should just have this be a date, not a date/time. It's not like something happens on the dot at the RSVP time, it's just a prod. And dealing with the time-picker is just another step, but if you don't then you get a random time assigned - the current time, which is likely not relevant. What do you think?

PresidentNick:
I was thinking we could use the RSVP by field to automatically lock the gig plans. We could default the time to midnight, but leave it available to  adjust if there is a use case for it. But if there is not use case, then we can just remove it.

PR #705, opened 2026-02-06
Add member API with creation endpoint

PresidentNick:
Blowcomotion wants to be able to create gig-o-matic members automatically now that we are signing up members on our band through a member signup form like this:

image

We can add password and password confirm fields to our member signup form to be able to send the minimum payload to create a user at the `/api/members` POST endpoint:

```json
{
  "email": "newuser@example.com",
  "username": "New User",
  "password": "SecurePassword123!"
}
```

creates:

image

Full Example with Optional Fields:

```json
{
  "email": "newuser1@example.com",
  "username": "New User",
  "password": "SecurePassword123!",
  "phone": "123456",
  "nickname": "noob"
}
```

creates:

image

We can then use the existing POST endpoints `band/assoc/[ID]/confirm` to add them to our band

and `/band/assoc/[ID]/tfparam` with `is_occasional: "true"` if our implemented attendance system hasn't marked them present in 90 days.

### DIFF
--- a/member/api.py
+++ b/member/api.py
@@ -2,6 +2,7 @@ from typing import Optional
 
 from django.http import JsonResponse
 from ninja import Router, Schema
+from ninja.throttling import AnonRateThrottle
 
 from member.models import Member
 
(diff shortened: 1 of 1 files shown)


PresidentNick:
This way (this PR) is not simpler! Not sending a password is better. I thought I was going to have to create the email. I can just reuse the signup view!

PR #708, opened 2026-02-09
Add member signup API and related functionality

PresidentNick:
- Implemented a new API endpoint for member signup.
- Created helper function to manage signup invitations.
- Updated views to utilize the helper function.
- Added tests for the new functionality.

### DIFF
--- a/member/helpers.py
+++ b/member/helpers.py
@@ -21,7 +21,7 @@ from django.contrib.auth.decorators import login_required
 from django.contrib.auth import logout
 from django.shortcuts import get_object_or_404, redirect, render
 from django.conf import settings
-from django.core.exceptions import ValidationError
+from django.core.exceptions import PermissionDenied, ValidationError
 from django.core.validators import validate_email
 from django.utils.translation import get_language_from_request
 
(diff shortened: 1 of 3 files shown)


aaronoppenheimer:
Just because the email exists in the system shouldn't cause a failure - maybe they're in another band (or already in this band). Also, should validate that the email address is a legit email address. If you haven't, look at member.views.InviteView which is what handles the "invite members" button in the band detail page.

PresidentNick:
@aaronoppenheimer I added the band invite API and kept the signup API as well, so both are available

PresidentNick:
@aaronoppenheimer yeah, two API endpoints here that are just API versions of these two forms:

https://gig-o-matic.com/member/signup

https://gig-o-matic.com/member/invite/[id]

I made the first API from the first url before realizing that the invite form for a band admin was also available (I hadn't really looked into the band admin that much). So I made another API version of the invite form. Instead of deleting the first one, I left it there in case someone might have a use case for that. I will definitely use the invite API for the band member signup on our website since it both signs up the member for gig-o-matic (if they don't have an account) and associates them with our band (also does that if they have an account). It just mimics the invite form that exists in the band admin.

PresidentNick:
@aaronoppenheimer What if we added throttling to the endpoints? I added the built-in django-ninja throttling with some reasonable limits.

PresidentNick:
@aaronoppenheimer Also, the APIs are still protected by requiring `X-API-KEY` auth, so the exposure to spam should be low, especially with added throttling.

aaronoppenheimer:
Is the "signup" API protected by the x-api-key? I see how the member invite code does it, but not the sign-me-up api. It wouldn't make sense anyway, would it? If the "sign me up" member had an API key?

PresidentNick:
@aaronoppenheimer I made those changes :D

PR #712, opened 2026-02-14
Clarified address label in GigForm to specify 'Street Address or Map URL

PresidentNick:
A bandmate who creates gigs thought "Address or URL" in the create gig form was for a web address, so they weren't filling out the address for gigs using that field. They were writing in the description. I changed the label for added clarity, since others might get confused too.


aaronoppenheimer:
Thanks for doing this - you're right, totally confusing. But it doesn't seem to be working in other languages - I still see the English version when I switch to French or whatever. The issue is the po file - if you search for "Street Address or Map URL" in there, you'll see it's marked with a comment "#, fuzzy" - this means (I think) that the translation system had an issue and won't actually map the msgid to the msgstr. If you remove that comment (and the one starting with #| if it's there - the old messageid) and recompile it should work.

Issue #646, opened 2025-07-10
Explain "simple planning"


aaronoppenheimer:
re #645 - should provide some explanatory text around the "simple planning" option so users know what it means!

PR #732, opened 2026-04-14
Make "Simple planning" option label text much more descriptive

kjcole:
Close issue #646

### DIFF
--- a/templates/band/band_form.html
+++ b/templates/band/band_form.html
@@ -145,7 +145,26 @@
                     <div class="form-group">
                         {{ form.simple_planning.errors }}
                         {% render_field form.simple_planning %}
-                        {{ form.simple_planning.label_tag }}
+                        <label for="{{ form.simple_planning.id_for_label }}"
+                               id="simple_planning_label">
+                            Check to disable 'Probably', 'Probably Not' and `Not Interested' attendance options.
+                        </label>
+
+                        <script>
+                          const cb     = document.getElementById("{{ form.simple_planning.id_for_label }}");
+                          const label  = document.getElementById("simple_planning_label");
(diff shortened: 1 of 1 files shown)


kjcole:
(It's dynamic too.)

aaronoppenheimer:
Cool! I'll check it out. We'll need to put in the translations too - I can do that if you don't want to take that piece on.

kjcole:
> Cool! I'll check it out. We'll need to put in the translations too - I can do that if you don't want to take that piece on.

I should have considered that... I'm guessing my "solution" isn't super elegant and may not stand up to adding lots of translations in a well-structured way, but it may give better minds an idea.

If wedging JavaScript in rather than taking advantage of Django's templating system isn't a great idea, perhaps a more robust but less interactive solution would be changing the "`Simple planning`" column / field to "`Expanded attendance options`" (or similar: "`Extended set`", "`Full set`", etc.) and reversing the logic on the checkbox: When checked (the default), the full set of attendance options is available. When unchecked the reduced set of attendance options is available. Then go back to letting Django's templating system do the heavy lifting.

aaronoppenheimer:
It might be even simpler just to list out the two sets of options and a pair of radio buttons. Just call it "attendance options" or something and present the two lists..?

aaronoppenheimer:
I changed the text to just "use yes/no/maybe responses" - is that clearer?

kjcole:
> I changed the text to just "use yes/no/maybe responses" - is that clearer?

Yes... but, perhaps indicate what that is in contrast to. Perhaps something akin to:

```
 Use Yes/Maybe/No Responses (eliminate "probably" and "probably not")
```

aaronoppenheimer:
Seems clunky. I'm going to leave it this way for now - it's better what was there before. I'll think about something better.

aaronoppenheimer:
(I've converted the PR to draft until there's a longer-term plan.)

PR #765, opened 2026-07-12
django 6

aaronoppenheimer:
Moved to Django 6. Requires Python 3.14 (or at least a more modern version than what we were using)
- **updated all packages**
- **updated python**


aaronoppenheimer:
@bklang So...this is gig-o ported to Django 6. I decided to leave the UI as-is - there are breaking changes and some things I don't understand about the later version of bootstrap. There's no code to change in porting; I turned off a linting error, updated all the libraries in requirement.txt, and changed the environment to Python 3.14 - it needed to be more recent than whatever I was running before. That'll mean updating the python on digitalocean manually, right? Or does that happen as part of pushing it out?

So - give this a try. All the tests pass, it looks like it's working...

PR #781, opened 2026-09-11
Fix duplicated gigs keeping original status and losing times

PresidentNick:
## Summary
- Duplicating a gig previously copied the original gig's status verbatim (e.g. a confirmed gig stayed confirmed), while the call/set/end times were dropped and defaulted to blank on the duplicate form.
- A duplicated gig is a new gig, so it should start Unconfirmed — but it's most likely happening at the same time as the original, so the times should carry over.
- `DuplicateView.get_form_kwargs` now overrides the copied `status` to `GigStatusChoices.UNCONFIRMED`.
- Added `DuplicateView.get_context_data` to populate `call_date`/`set_time`/`end_date` from the original gig, the same context variables `UpdateView` already provides for its date/time picker widgets — `CreateView` (which `DuplicateView` extends) never set these, which is why the times appeared erased.

## Test plan
- [x] `gig` app test suite (96 tests, including `test_duplicate_gig`) passes locally
- [x] Manually duplicated a gig locally and confirmed the new gig form shows Unconfirmed status with the original's times pre-filled

### DIFF
--- a/gig/tests.py
+++ b/gig/tests.py
@@ -946,13 +946,17 @@ class GigTest(GigTestBase):
         number=1,
         user=None,
         expect_code=302,
-        call_date="01/02/2100",
-        end_date="",
-        call_time="12:00 pm",
-        set_time="",
-        end_time="",
         **kwargs,
     ):
+        # like update_gig_form, default the date/time fields from the gig being
+        # duplicated - this mirrors what the duplicate form pre-fills in the browser
+        call_date = kwargs.pop("call_date", self._dateformat(gig.date))
+        end_date = kwargs.pop("end_date", self._dateformat(gig.enddate))
+        call_time = kwargs.pop("call_time", self._timeformat(gig.date))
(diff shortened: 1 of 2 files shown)


PresidentNick:
This one was requested by @blowcomotionband

aaronoppenheimer:
Please add a unit test for this case. There are already some tests that duplicate gigs - I think if you copy the one that uses the form and leave out the date/time fields, they'll be copied over from the original, which you can check.

PresidentNick:
Added `test_duplicate_gig_copies_times_and_resets_status`. I also updated `duplicate_gig_form` to default the date/time POST fields from the original gig (mirroring how `update_gig_form` defaults from `the_gig`), so leaving those fields out of a test call simulates what the pre-filled duplicate form actually submits. The new test confirms the duplicated gig keeps the original's call/set/end times while its status resets to Unconfirmed.

aaronoppenheimer:
Looks good! Thanks!

PR #784, opened 2026-09-16
Fix for schedule filters not working

jaswope:
#780 upgraded to htmx4 but left a few issues. Primarily the filters on the schedule page stopped working.

This PR introduces a few changes that fix that issue and clean up some upgrade issues:
- The reload, reload-page, and dropdown extensions have been reimplemented in idiomatic htmx4
- The above extensions are now referenced the correct way on elements with an attribute (the old `hx-ext` method has been fully deprecated)
- The checkbox extension has been removed (#783 made it obsolete)


aaronoppenheimer:
THANK YOU!

Issue #761, opened 2026-06-22
Filter for the schedule page to turn off canceled & can't-do-it gigs


aaronoppenheimer:
An option to just show the gigs I've said I can do (or haven't weighed in) - not gigs I can't do or have been canceled.

PR #787, opened 2026-09-18
Add a button to hide cancelled and declined gigs

jaswope:
This addresses #761. I struggled with the wording, but decided on "Not Canceled or Declined". Happy to change it. I opted to exclude "Cancelled", "Not Interested", and "Can't Do It". I left any other response (including "Possibly Can't").

I also included a small refactor: The same logic for getting the clean list of gigs was duplicated several times and my change would have added yet another. I instead moved it alongside "future_plans" and "future_noplans" to keep things DRY and organized.

I have not included translations (they would have been auto-generated in any case if I did). Let me know if you want that.


agenda/views.py
```diff
@@ -96,6 +96,8 @@ def get_context_data(self, **kwargs):
                  layout==AgendaLayoutChoices.ONE_LIST, False],
                 [AgendaLayoutChoices.NEED_RESPONSE, _('Needs Response'), 0,
                     layout==AgendaLayoutChoices.NEED_RESPONSE, True],
+                [AgendaLayoutChoices.HIDE_DECLINED_AND_CANCELED, AgendaLayoutChoices.HIDE_DECLINED_AND_CANCELED.label, 0,
```
jaswope, line 99: I did this slightly different from the existing pattern to prevent duplicating the same string around. It should work exactly the same WRT to translations though. It may be worth taking a look to see if we want to give the same treatment to the other buttons.

aaronoppenheimer:
@jaswope I think it'd be better to do this as a toggling button like "show locations" - if you're a member of multiple bands you get more tabs so you can see each band's gigs individually. If you're someone who likes to do that and you want to hide declined gigs, this solution gets complicated. Doing it like "show locations" makes it a global setting that would apply to all of the panels.

Maybe both "show location" and "hide declined/canceled" should be checkboxes instead of a button with toggling label, too.

jaswope:
Ahh, I considered that approach, but decided not to do that (and was going to recommend making these more of a "filter") because I didn't see any prior patterns there. I'll take another look, and I'm guessing it'll likely be a different PR rather than just changing this one. I'll look to do it as a second button for now, and then a UI change can be follow-on.

jaswope:
@aaronoppenheimer OK, it's changed to the button style instead. Decided to leave it in the same PR because some of the refactor would still be nice to keep around. I was considering adding a badge to show many are hidden/would be hidden by toggling it. What do you think?

aaronoppenheimer:
I think that's a good idea. Consider the case where my band has 10 upcoming gigs, and I have declined 5 of them. In that case, the tab at the top says "Upcoming Gigs (10)" but only 5 are shown. It's better that the declined filter button says "Show Canceled/Declined (5)" than changing the tab or promising 10 but only showing 5 without comment.

Also, if I decline *all* of my band's gigs, the agenda says "No gigs!" but could say "All gigs canceled/declined" if that's why there aren't any.

PR #788, opened 2026-09-19
Timezone and DST fixes

jaswope:
While working on a much larger recommendation for handling gigs in different timezones, I wanted to fix a bunch of outstanding timezone and DST handling issues first. This is a large PR, but I opted for that instead because splitting it up would have caused conflicts and these are all somewhat related. If preferred I can break these down, and the breakdown will approximately be per commit in this branch (a few would be combined). Looking at the PR on a commit by commit basis will probably be the best way to review it in small chunks.

### Upgrading babel and pinning tzdata
General work needed before other TZ work. The babel work is mostly for a later PR about gig time rendering, but it's worth keeping it up to date.
### Validate and refactor user timezone handling
Timezones are now validated on load and store. User timezone is determined in the middleware only now, and the one determined there (with UTC as the de facto fallback if no zone is configured or the loaded zone is invalid) is used for the rest of the UI instead of reparsing it. If an unsupported timezone change is detected, we don't update it.

This also removes code from agenda.html that was related to setting the user's timezone but never actually reachable (`current_timezone` was never defined in the context) and could have resulted in redirect loops.
### Email related timezone bugs
A user with a bad timezone will now not throw an exception during email building, and we will fall back to using the band's timezone. Additionally, using `override` instead of `activate` to prevent the member's tz from leaking out into the entire thread's TZ rendering.
### Full-day event timezone fix
Previously, full day events for some timezones incorrectly appeared to cross day boundaries. Specifically for timezones east of UTC.

One note, the test for this requires some juggling of the band timezone in order to get things back to a valid test state due to memoization of the band object. The actual risk of encountering this bug is minor (it would require someone to update the band info at the exact same moment as a gig is edited/created) but the fix is more scope than I want in this fix.
### Gig heatmap fixes
The gig heatmap is now calculated completely in the band timezone. Previously it was calculating in (essentially) UTC resulting gigs crossing day/year boundaries incorrectly and causing an inaccurate heatmap.
### DST bug on gig series
When a series of gigs is defined, we should schedule them for the same wall clock time honoring the DST changes for the applicable time zone. Previously we were using the same offset for the series, resulting in gigs being off by and hour after a DST boundary was passed.
### Invalid or ambiguous time crash in gig form due to DST
The gig forms now returns a validation error if a time is chosen that does not exist, and chooses the first occurence if it is ambiguous

Previously the form would crash on these two cases. 

A note: This implementation is not the ideal solution (I do not like passing the field name through to `_to_band_time`) but extra tests are added to prevent regressions. An ideal solution will require a much larger refactor of the form handling code, and the tests can be removed then. I plan to have a PR later that should address this.
### Snooze feature used server (UTC) timezone instead of user timezone
This one is really minor, but would result in the business logic for determining when a reminder email should be sent to be off by a day, which somewhat matters when the shortest time is 2 days.


aaronoppenheimer:
(I added and deleted a comment about a bug -> user error)

aaronoppenheimer:
This is great - thanks for taking this on. Timezones have caused more problems in the gig-o than anything else.

